PR #99, opened 2017-04-07
Rules for Shield Controller

R3d-Sky:
See libretro/Lakka#379. Update udev rules to use the pre-2017 Shield Controller.


Kivutar:
Thanks

PR #101, opened 2017-04-07
Create 99-gen-game-s3-controller.rules

hooby3dfx:
Another controller that also brings along a keyboard and mouse profile


Kivutar:
Thanks!

PR #102, opened 2017-04-07
Update platform_init

NuisMons:
Ignores maxRes to set in framebuffer


Kivutar:
Thanks, I will try it on my setup.

Issue #113, opened 2017-04-27
Compiling Lakka for Cubox fails @ Building Retroarch


thefoofighter:
Hi,

I tried posting on the forums and either I am too thick to find the create thread button or there is some kind of time limit preventing new users from creating threads. Irregardless it might make more sense here anyway.

Background:
I own a cubox-i 4 pro which I recently loaded LibreELEC on after switching from OpenELEC. The latest builds using kernel 4.4 cause a black screen on boot for some Cubox-i owners. Vpeter over at LibreELEC has kindly built the latest sources for the community using the 3.14 Kernel which thankfully works on my device.

in this Lakka repo for LibreELEC he has also added the kernel sr-3.14 for solidrun devices that have this issue. 

I attempted to build the sources using a clean Linux Mint installation for it to fail it seems at the retroarch build.

If this is not the right place to post this feel free to close it but please point me in the right direction :)

Here is the log. I ran this as a regular user.

```
thefoofighter@LinuxMint ~/Dev/Lakka-LibreELEC $ DISTRO=Lakka PROJECT=imx6 SYSTEM=cuboxi ARCH=arm LINUX_VERSION=sr-3.14 make image
./scripts/image mkimage
 =================================================================================
 Configuration for Lakka (community)
 =================================================================================

 Buildoptions:
 ======================================================
 - CPU (ARCH):				 cortex-a9 (arm)
 - FLOAT:				 hard
 - FPU:					 neon
 - SIMD support:			 yes
 - LTO (Link Time Optimization) support: yes
 - GOLD (Google Linker) Support:	 yes
 - LLVM support:			 

 Graphic configuration:
 ======================================================
 - OpenGL (GLX) support (provider):	 no (no)
 - OpenGLES support (provider):		 yes (imx-gpu-viv)
 - uvesafb support:			 no

 Hardware decoder configuration:
 ======================================================
 - Kodi Player driver:			 libfslvpuwrap
 - VAAPI Support:			 
 - VDPAU Support:			 

 Input device configuration:
 ======================================================
 - Remote support:			 no
 - ATV Remote support:			 no
 - CEC Adapter support:			 yes
 - Kodi Joystick support:		 yes

 Misc. hardware configuration:
 ======================================================
 - ALSA support:			 yes
 - Pulseaudio support:			 no
 - Bluetooth support:			 yes
 - Include driver:			 RTL8192CU
 - Include driver:			 RTL8192DU
 - Include driver:			 RTL8192EU
[...]

PR #114, opened 2017-04-28
Fix for Issue 113 [Compiling Lakka for Cubox fails @ Building Retroarch]

thefoofighter:
I dug a little deeper and I found a solution to issue #113
packages/libretro/retroarch/package.mk was missing the correct $OPENGLES param for the GPU on a Cubox.
"imx-gpu-viv" is the correct string in this case.
Maybe there should be a check here that informs someone that $OPENGLES is set to an unknown parameter?


Kivutar:
Thank you

PR #149, opened 2017-06-18
Update citra and fix compilation issue

gouchi:


Kivutar:
Yey! Can't wait to test it :D

PR #175, opened 2017-07-13
Update core-info

### LINKED ISSUE #173: [request] bundle new MAME 0.78 playlist scanner database in next release
After some wrangling, there are now playlist scanner databases for each of the frozen MAME cores up to MAME 2014.

Would you be willing to consider bundling at least the MAME 2003/0.78 RDB in the next Lakka release, as MAME 2003 is part of the standard packages of cores?
https://github.com/libretro/libretro-database/blob/master/rdb/MAME%200.78.rdb

If you were to bundle a second RDB in Lakka, I'd humbly suggest MAME 2014/0.159 even though you aren't including that core in the Lakka images.
https://github.com/libretro/libretro-database/blob/master/rdb/MAME%200.159.rdb

P.S. Each of those RDBs also has a corresponding .info file in the libretro-super repository.

RobLoach:
Update core-info to latest version from libretro-super.

Fixes #173

### DIFF
--- a/packages/libretro/core-info/package.mk
+++ b/packages/libretro/core-info/package.mk
@@ -19,7 +19,7 @@
 ################################################################################
 
 PKG_NAME="core-info"
-PKG_VERSION="e50f14a"
+PKG_VERSION="665c5f4"
 PKG_REV="1"
 PKG_ARCH="any"
 PKG_LICENSE="GPL"


markwkidd:
Thanks for the quick attention @Kivutar and @RobLoach . Does the PR add the RDB file or only the .info file? From what I can see it's only the .info file

RobLoach:
RDBs at https://github.com/libretro/Lakka-LibreELEC/pull/174

markwkidd:
Ah, now I see. Thanks @RobLoach

PR #209, opened 2017-09-06
Fixed up noobs slides and sources

numanair:
Added correct source SVGs for noobs slides and updated the wording on them. Some of the wording could still be better, but this fixes typos and the biggest issues with it.


Kivutar:
Thank you!

PR #256, opened 2017-10-23
Update glib to 2.54 to address compile errors

Zerauskire:
Update glib to address compile errors in newer versions of GCC.

### DIFF
--- a/packages/devel/glib/package.mk
+++ b/packages/devel/glib/package.mk
@@ -18,11 +18,11 @@
 ################################################################################
 
 PKG_NAME="glib"
-PKG_VERSION="2.46.2"
+PKG_VERSION="2.54.1"
 PKG_ARCH="any"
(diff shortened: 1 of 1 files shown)


Kivutar:
We can keep this PR open for documentation purposes. But I can't merge it at this stage of the release process. Also I like to stick to LibreELEC when possible.

Zerauskire:
Ok. No problem. LibreELEC is currently on 2.53.5 which would address this issue as well. So maybe just pull down from LibreELEC?

https://github.com/LibreELEC/LibreELEC.tv/blob/master/packages/devel/glib/package.mk

Zerauskire:
Current version of LibreELEC also has an update to automake that addresses a regex issue i'm running into while compiling.

Kivutar:
This branch is based on the stable LibreELEC 8.2
And the next version will be based on 8.4 or the next one.

PR #266, opened 2017-10-31
"hack" to avoid rebuild whole build. dirs in case of version change.

ToKe79:
Attention: the original build. dirs must be renamed to folders without
the suffix, e.g. rename build.Lakka-Generic.x86_64-8.2-devel to
build.Lakka-Generic.x86_64
-vudiq


Kivutar:
Thanks for the tip. Too hacky to be merged though.

PR #298, opened 2017-11-29
update libretro cores and retroarch

ToKe79:
redream and sameboy needed additional changes / discussed with @Ntemis via IRC

### DIFF
--- a/packages/libretro/sameboy/package.mk
+++ b/packages/libretro/sameboy/package.mk
@@ -19,7 +19,7 @@
 ################################################################################
 
 PKG_NAME="sameboy"
-PKG_VERSION="8d69156"
+PKG_VERSION="8c9ff67"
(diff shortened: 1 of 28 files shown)


packages/libretro/sameboy/package.mk
```diff
@@ -35,10 +35,11 @@ PKG_IS_ADDON="no"
 PKG_AUTORECONF="no"
 
 make_target() {
-  make -f Makefile.libretro
+  cd $PKG_BUILD/libretro
+  make -f Makefile
```
RobLoach: Could likely use....
```
make_target() {
  make -C libretro
}
```

Save yourself having to change directory yourself.

Ntemis: Will merge and fix it

PR #302, opened 2017-12-07
New script to update only used libretro packages

ToKe79:
This script is used to build daily with latest versions of libretro packages.


RobLoach:
Looks quite similar to the other script. Could it become an option that you pass in to save code duplication?

```
./libretro_update.sh --used
```

ToKe79:
Now it accepts one parameter - either --all (-a) or --used (-u). Without parameter it prints out the usage.

RobLoach:
Nice work! @Ntemis, @Kivutar, what do you think?

PR #306, opened 2017-12-09
Fixes issue #283 - driver for RTL8192EU

ToKe79:
added wifi driver for the OdroidXU3 project to fix issue #283 
maybe also other wifi drivers could be added

### DIFF
--- a/packages/linux-drivers/RTL8192EU_43/_patches/RTL8192EU-gcc-5.patch
+++ b/packages/linux-drivers/RTL8192EU_43/_patches/RTL8192EU-gcc-5.patch
@@ -1,0 +1,25 @@
+diff -Naur a/include/ieee80211.h b/include/ieee80211.h
+--- a/include/ieee80211.h
++++ b/include/ieee80211.h
+@@ -1314,18 +1314,18 @@
+ (((Addr[2]) & 0xff) == 0xff) && (((Addr[3]) & 0xff) == 0xff) && (((Addr[4]) & 0xff) == 0xff) && \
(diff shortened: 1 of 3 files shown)


Ntemis:
You didn't get a confirmation from the user, did you test it locally?

Kivutar:
Thank @ToKe79 
Did you try that change?

ToKe79:
TL;DR; Don't merge.

Had to use driver based on 4.3 in the end and [confirmed](https://github.com/libretro/Lakka-LibreELEC/issues/283#issuecomment-352030576) working by the user. So as both the 4.4 driver and also 4.3 driver have same kernel module - 8192eu.so - they cannot be used at the same time, but I managed to create a RTL8192EU_43 package and build with that for the user.
So the choices are:
- use the 4.3 driver (were there any reports for 2.0 stable that this driver was not working for somebody?)
- use the 4.4 driver (and have some users without working wifi)
- open an issue at Mange/rtl8192eu-linux-driver and report regression on the TP Link WN823N v2 with the 4.4 based driver

However this PR is obsolete and should not be merged - first I thought it helped the user (he was very [grateful](https://github.com/libretro/Lakka-LibreELEC/issues/283#issuecomment-350432543) for the first image), asked and waited for confirmation, then after a day without an answer made the PR. Later he reported that it did not fix the issue so investigated further until I found out that Lakka 2.0 used a different version of the driver.

ToKe79:
I pushed the final changes, so you see how i fixed the issue.

PR #311, opened 2017-12-13
libretro-database: Update to the latest database

RobLoach:
This updates to the latest libretro-database, with a recent RDB update.


Ntemis:
just updated all cores.

PR #328, opened 2018-01-08
Various build fixes

bkoropoff:
This fixes build problems I ran into on Fedora 27,


bkoropoff:
Added some unpacking changes to save bandwidth

Kivutar:
Your idea of using PKG_GIT_URL instead of the libretro PKG_SECTION is good.

Where do the other changes to patches come from?

bkoropoff:
@Kivutar I just wrote the patches by working through the build errors one at a time.

ToKe79:
`PKG_GIT_URL` should be then also added to `config/path`, i.e. set it to empty, so a value from previous package is not re-used with current package if not set in current `package.mk` (in the section after `# set package metadata`). But maybe there is a different reason why all the PKG_* are re-set in config/path.

bkoropoff:
@ToKe79 done

packages/libretro/sameboy/package.mk
```diff
@@ -25,6 +25,7 @@ PKG_ARCH="any"
 PKG_LICENSE="MIT"
 PKG_SITE="https://github.com/libretro/sameboy"
 PKG_URL="https://github.com/libretro/sameboy/archive/$PKG_VERSION.tar.gz"
+PKG_SOURCE_DIR="SameBoy-$PKG_VERSION*"
```
RobLoach: * ?

bkoropoff: If you don't specify the name of the directory that gets unpacked from the tarball, it guesses (in this case incorrectly) from the package name.

PR #332, opened 2018-01-14
Add RPi3 version

bkoropoff:
This is basically a clone of RPi2 except the compiler flags are optimized for the CPU/FPU in the Pi 3.

The only interesting part is:

```sh
        TARGET_CPU="cortex-a53"
        ...
        TARGET_FPU="neon-fp-armv8"
```

I'm not sure if this is anything Lakka cares to support/maintain, but I've been using it successfully for private builds.


kivutar:
Thanks for the PR.

I don't think that the performance gain is worth the maintenance cost.

PR #333, opened 2018-01-14
fix: sources/unpack - when using PKG_GIT_URL

ToKe79:
when using PKG_GIT_URL and no PKG_URL, scripts/get exits with 0 (PKG_URL is empty) and hence no $SOURCES/$1 folder exists. So enhancing the condition in scripts/build with PKG_GIT_URL check - if not empty, continue with the unpack.
Please test, my daily builds are running, cannot test on my side.
Affected citra when building from clean repository. This should fix it.

### DIFF
--- a/scripts/unpack
+++ b/scripts/unpack
@@ -60,7 +60,7 @@ fi
 
 [ -f "$STAMP" ] && exit 0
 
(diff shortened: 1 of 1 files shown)


Ntemis:
Confirmed, thank for the quick fix

PR #334, opened 2018-01-15
more fixes due to PKG_GIT_URL

ToKe79:
previously git clones were made in case of PKG_SECTION == libretro, now it is based on PKG_GIT_URL (PR 328). but many libretro packages were not modified to have PKG_GIT_URL and hence broke - could not be correctly downloaded / cloned / built.

I have now added PKG_GIT_URL, removed the PKG_URL and also removed the post_unpack() function in most of the libretro packages (all except psp1 - seems to be obsolete).

It is most certain that these updates of the package.mk files will trigger rebuilding of the libretro packages (one timer).


bkoropoff:
@ToKe79 The point of the change was to avoid git cloning a repo on every rebuild.  I switched packages that have submodules to `PKG_GIT_URL` since the archive tarballs did not include them.  The others (at least the ones in the RPi3 target) built fine from tarball.  Is there a technical reason to always use git clones for libretro packages?

ToKe79:
Hi,
yes, besides the submodules reason, the git version is compiled into the core and compared for netplay - they must be same.

bkoropoff:
@ToKe79 Any way we could pass the git version from the package.mk into the build?  Some of the git clones are enormous and could be a problem for people with data capped internet, etc.

bkoropoff:
@ToKe79 Another thought: if git clones are unavoidable, maybe clone into the sources directory on the first build, and thereafter symlink it into the build directory and do a `git clean` and `git checkout` on package rebuilds.  That would at least avoid a fresh clone on every package update.

ToKe79:
But the repos are cloned / packages rebuilt only in case a package.mk changed. Or should at least and also behaves like that here.

bkoropoff:
@ToKe79 Sure, but if you are staying on top of retroarch master with libretro_update.sh, that can be quite often.

ToKe79:
git clone retroarch repo: 166 MB
wget master.tar.gz: 20 MB
if the bandwidth is the reason, then initial git clone into sources/ with subsequent git clean / git pull / git checkout version should save lot of bandwidth.
I will look into that.

ToKe79:
So here is what am I thinking of currently:

in `scripts/unpack` replace part between lines 105-114 with:
```bash
if [ -z "$PKG_URL" -a -n "$PKG_GIT_URL" ]; then
  if [ -d "$SOURCES/GITS/$1" ]; then
    cd $SOURCES/GITS/$1
    git clean -fx
    git checkout -- .
    git pull
  else
    if [ ! -d "$SOURCES/GITS" ]; then
      mkdir -p "$SOURCES/GITS"
    fi
    git clone $PKG_GIT_URL $SOURCES/GITS/$1
    cd $SOURCES/GITS/$1
  fi
  git checkout $PKG_VERSION
  git submodule update --init --recursive
  ln -s $SOURCES/GITS/$1 $BUILD/$1-$PKG_VERSION
  cd $ROOT
elif [ -z "$PKG_URL" ]; then
  mkdir -p "${BUILD}/${PKG_NAME}-${PKG_VERSION}"
fi
```
in `scripts/clean`: when removing sources (`$CLEAN_SOURCES = true`) add condition that in case `PKG_GIT_URL` is not empty then remove from `$SOURCES/GITS/$1` (instead of `$SOURCES/$1`). The `rm -rf $i` should remove only the symbolic link, not the linked `$SOURCES/GITS/$1`.

@Kivutar @Ntemis @gouchi @RobLoach - what do you think? I can first test it with nightlies on my build host (after current run finishes), but if it is the correct direction/way of doing this.

ToKe79:
FYI: commits up to now in this PR are tested!

Ntemis:
Will review your changes today. Thanks

Kivutar:
I took a quick look at the commit diff, but it's difficult to understand the changes in scripts/ with all the indentation changes.

ToKe79:
actually the only changes in `scripts/unpack` is added `if [ -z "$PKG_GIT_URL" ]` (currently on line 76) and `fi` (currently on line 103), hence the change of the indentation. This should skip all the pre_unpack, unpack, etc. as it is not needed when doing a git clone. This avoids unpacking previously downloaded sources from git in `$SOURCES/$PKG_NAME` as tar.gz archives - but only in case PKG_GIT_URL is set, which is currently set only in libretro packages.

ToKe79:
actually, the idea with symbolic links is a NO-GO:
in $PKG_BUILD/.install_pkg the target files from last PROJECT will remain and in case no new rebuild is necessary, these files would be copied to $BUILD/image/system and we would end up with binaries of e.g. Rockchip.MiQi.arm in image for e.g. Generic.x86_64.

ToKe79:
different idea:

```bash
if [ -z "$PKG_URL" -a -n "$PKG_GIT_URL" ]; then
  # check if we have existing a repository
  if [ -d "$BUILD/$1-*" ]; then
    # rename original dir to new version
    mv $BUILD/$1-* $BUILD/$1-$PKG_VERSION
    cd $BUILD/$1-$PKG_VERSION
    # remove all changes
    git clean -fx
    git checkout -- .
    # update repository
    git pull
  else
    # clone repository into build
    git clone $PKG_GIT_URL $BUILD/$1-$PKG_VERSION
    cd $BUILD/$1-$PKG_VERSION
  fi
  git checkout $PKG_VERSION
  git submodule update --init --recursive
  cd $ROOT
elif [ -z "$PKG_URL" ]; then
  mkdir -p "${BUILD}/${PKG_NAME}-${PKG_VERSION}"
fi
```
There is one "quirk" - all libretro packages which have a dash (`-`) in their name must be renamed, so e.g. retroarch-assets-* is not matched when building retroarch-*. So we could replace the `-` with `_` in package names. Don't the current package names conflict now when doing `scripts/clean retroarch`?

This should save bandwidth and also time when bumping the version - we do not have to reclone the whole repository, only pull the changes.

If it will get accepted, then probably out of scope of this PR?

Ntemis:
GTM @kivutar

PR #343, opened 2018-01-24
Generic: move additional libretro cores

ToKe79:
move additional libretro cores from `packages/lakka/Generic/package.mk` to `projects/Generic/options`, where it belongs


kivutar:
Good catch.

Can you do this instead?

 - Adding these cores to distributions/Lakka/options
 - In the package.mk of these cores, set PKG_ARCH="x86_64 i386"

It will have the same effect, and it's even more generic.

ToKe79:
done, i will update the libretro_update.sh script, currently those cores are added manually, so they would be checked twice.

PR #347, opened 2018-01-25
gearboy: bump version to latest, 6b69f16 does not exist

ToKe79:
Tested only build, not the core behavior / urgent fix for being able to build.


RobLoach:
Hmm? https://github.com/libretro/Gearboy/commit/6b69f16

ToKe79:
That is strange, `git checkout 6b69f16` throws an error during build.

ToKe79:
see issue #348 - same error as I got.

PR #352, opened 2018-01-27
Update Non-Commercial packages to nonfree

RobLoach:
Updates the non-commercial packages from [libretro license docs](https://docs.libretro.com/tech/licenses/) to `nonfree` to match some of the other packages in LibreELEC:
https://github.com/LibreELEC/LibreELEC.tv/blob/master/packages/addons/driver/sundtek-mediatv/package.mk#L23


RobLoach:
Going to change the LIBRETRO_CORES var instead.

PR #356, opened 2018-01-29
scripts/unpack: save bandwidth/time using git pull [WIP]

ToKe79:
For packages where we use a git repository instead of sources archive, we do not need to clone the whole repository every time.

At the beginning we must source the package.mk to have a valid PKG_GIT_URL to skip the cleaning of the old version. Then, instead of doing a `git clone` we:
- find the previous package build folder in BUILD (by comparing PKG_NAME with STAMP_PKG_NAME in .libreelec-unpack and then checking if PKG_GIT_URL is in .git/config) -- this confirms that we are in the correct folder AND the folder is a cloned repository
- remove all changes in the folder - `git clean -fdx` cleans all untracked files and folders and `git checkout -- .` removes all changes to original files (e.g. when patching some files)
- then we must switch to an existing branch - `git checkout master` so we can
- pull the changes from the GitHub repository - `git pull`

If such folder is not found, we use the clean script to clean any old build folders of that package and clone the repository.

We checkout the commit which we want to compile using `git checkout $PKG_VERSION` as previously (this is not changed). The GIT_VERSION for each individual libretro core is parsed in one of the Makefiles of each core. This is also not changed, so cores will build with the GIT_VERSION information.

This is first concept with some debug messages and no indentation changes for easier tracking of the changes.

### DIFF
--- a/scripts/unpack
+++ b/scripts/unpack
@@ -41,7 +41,7 @@ STAMP_DEPENDS="$PKG_DIR $PKG_NEED_UNPACK $PROJECT_DIR/$PROJECT/patches/$PKG_NAME
 
 # Perform a wildcard match on the package to ensure old versions are cleaned too
 PKG_DEEPMD5=
-GIT_MUST_REBUILD=false
+GIT_MUST_REBUILD="no"
 for i in $BUILD/$1-*; do
   if [ -d $i -a -f "$i/.libreelec-unpack" ] ; then
     . "$i/.libreelec-unpack"
@@ -51,7 +51,7 @@ for i in $BUILD/$1-*; do
         if [ -z "$PKG_GIT_URL" ] ; then
         $SCRIPTS/clean $1
         else
-          GIT_MUST_REBUILD=true
+          GIT_MUST_REBUILD="yes"
         fi
       fi
     fi
@@ -63,7 +63,7 @@ if [ -d "$PKG_BUILD" -a ! -f "$STAMP" ]; then
   $SCRIPTS/clean $1
 fi
 
-[ -f "$STAMP" -a !$GIT_MUST_REBUILD ] && exit 0
+[ -f "$STAMP" -a "$GIT_MUST_REBUILD" = "no" ] && exit 0
 
 if [ -d "$SOURCES/$1" -o -d "$PKG_DIR/sources" -o -n "$PKG_GIT_URL" ]; then
   printf "%${BUILD_INDENT}c ${boldcyan}UNPACK${endcolor}   $1\n" ' '>&$SILENT_OUT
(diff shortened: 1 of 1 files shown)


kivutar:
Sorry for the hard work you put on this @ToKe79 but I prefer to avoid huge changes to `scripts/` as it makes merging upstream hazardous.

Also, a more complex implementation like this is also more prone to errors. For example, sourcing package.mk can cause variable name collisions during an upstream merge.

I'm going to wait for the reviews of other contributors.

scripts/unpack
```diff
@@ -103,9 +108,55 @@ if [ -d "$SOURCES/$1" -o -d "$PKG_DIR/sources" -o -n "$PKG_GIT_URL" ]; then
...
+        if [ -f $f/.libreelec-unpack ]; then
+          echo "**** debug: '.libreelec-unpack' found"
+          source $f/.libreelec-unpack
+          echo "**** debug: STAMP_PKG_NAME=$STAMP_PKG_NAME"
+          if [ "$PKG_NAME" = "$STAMP_PKG_NAME" ]; then
+            echo "**** debug: NAME matched"
+            if [ -f $f/.git/config ]; then
+              echo "**** debug: .git/config found:"
+              cat $f/.git/config
+              GIT_MATCH=$(grep "$PKG_GIT_URL" $f/.git/config | wc -l)
```
RobLoach: Rather than parsing .git/config, you could likely use `git remote get-url origin`

scripts/unpack
```diff
@@ -103,9 +108,55 @@ if [ -d "$SOURCES/$1" -o -d "$PKG_DIR/sources" -o -n "$PKG_GIT_URL" ]; then
...
+      git checkout master
+      # pull changes
+      echo "**** debug: git pull"
+      git pull
+    else
+      # clean
+      $SCRIPTS/clean $1
+      # clone
     git clone $PKG_GIT_URL $BUILD/$PKG_NAME-$PKG_VERSION/
     cd $BUILD/$PKG_NAME-$PKG_VERSION/
```
RobLoach: Whitespace fix.

RobLoach:
While I like the general idea,  I'm with Kivutar. This seems like something you could push forwards to [LibreELEC](https://github.com/LibreELEC/LibreELEC.tv). I'm sure they'd love to see a sped up git clone process.

ToKe79:
Thank you for you input.

I parse `.git/config` due to the fact, that:
- I make sure that the current folder is a git repository
- and that the url is the same as in the package.mk
In case I would not check for .git/config and just do `git remote get-url origin` and the current folder would not be in a git repository, I could get the url of the Lakka git. But the result would be same - it would not match and I would clean :-)

I also understand that merging this to LE might be complicated (LE to my knowledge does not use `git clone` at all) and maintaining this change locally when merging from LE could and probably will introduce some overhead. But to my defense :-) the original code for `git clone` based on PKG_SECTION is also not merged upstream.

I have checked if sourcing package.mk is really necessary, but it is not, as is already done with this code:
https://github.com/libretro/Lakka-LibreELEC/blob/07b402ca29ce2c32d5df2ef20803f3aacf595a49/scripts/unpack#L21

I can change also the checking of the git url as @RobLoach pointed out - seems more cleaner way . And no hard feelings if this is not going to be merged.

scripts/unpack
ToKe79: On which line was the incorrect whitespace? I cannot see the change.

RobLoach: Ah, the `git clone` and `cd ...` should be two spaces indented to make sure the whitespace is correct.

Line 157 and 158

ToKe79:
There is git handler in LE9 which solves this.

PR #357, opened 2018-01-29
Build while omitting specific cores

ToKe79:


kivutar:
It would be easier to just do a concatenation instead of a loop no?

```
if [ -z $INCLUDE_NONFREE ] ; then
    LIBRETRO_CORES="$LIBRETRO_CORES $LIBRETRO_NC_CORES"
fi
```

ToKe79:
In LIBRETRO_NC_CORES are also cores, which are currently not in LIBRETRO_CORES (e.g. mame, mame2014, etc.), therefore I think it is more cleaner to keep all cores, which are built, in LIBRETRO_CORES and all non-commercial cores, for which a package exists (regardless if they are in LIBRETRO_CORES or not), in LIBRETRO_NC_CORES. But I could rename LAKKA_PARTNER_BUILD to EXCLUDE_NONCOMMERCIAL (but it is only semantics, as non-free is not same as non-commercial).

ToKe79:
not merging

kivutar:
Ok, I didn't get that LIBRETRO_NC_CORES had cores that weren't in LIBRETRO_CORES at the first read.

But I don't see the point of having these extra core listed in LIBRETRO_NC_CORES, if we don't use them.

RobLoach:
On the LibreELEC side, it looks like it does something with KODI_NONFREE_SUPPORT:
https://github.com/libretro/Lakka-LibreELEC/blob/8b8d1c994ae471042c0a0115d0253c81f041b147/packages/mediacenter/kodi/package.mk#L92-L96

Could adopt a similar pattern.

PR #365, opened 2018-02-01
XU[34]: Remove useless files

kivutar:
I'm able to build without these.
I took care of removing them from the build folder before testing, and I did a scripts/clean retroarch


kivutar:
Tested on the XU4

PR #367, opened 2018-02-01
tests: Add very basic tests in bash

kivutar:


RobLoach:
Would this work on Jenkins?

ToKe79:
If we would add it before the `make` command, we could get an error and exit with error without starting the build, so I think yes.

PR #372, opened 2018-02-04
XU[34] Fix display of the partition resize messages

kivutar:
* console=tty0 needed to be after the other console=
 * quiet is needed to not show the systemd boot log when tty0 works
 * For this particular device, retroarch can work without the mali module being loaded, and loading this module was failing anyway


Ntemis:
Looks good. Testing ram freq at 866 vs 933 if all works ok better to set that as default.

PR #377, opened 2018-02-07
XU[34]: Set memory freq at 1800Mhz

Ntemis:


Ntemis:
1866Mhz was causing me some lags with this all looks good

Issue #378, opened 2018-02-08
Python3 required for vulkan-loader [Tinkerboard]


ToKe79:
Ran into build error while compiling the Lakka-2.1-dev branch on a minimal system (Debian 9.3):
```
  BUILD    vulkan-loader (target)
...[redacted]...
CMake Error at /home/vudiq/projects/Lakka-LibreELEC/build.Lakka-TinkerBoard.arm-2.1-devel/toolchain/share/cmake-3.6/Modules/FindPackageHandleStandardArgs.cmake:148 (message):
  Could NOT find PythonInterp: Found unsuitable version "2.7.13", but
  required is at least "3" (found
  /home/vudiq/projects/Lakka-LibreELEC/build.Lakka-TinkerBoard.arm-2.1-devel/toolchain/bin/python)
Call Stack (most recent call first):
  /home/vudiq/projects/Lakka-LibreELEC/build.Lakka-TinkerBoard.arm-2.1-devel/toolchain/share/cmake-3.6/Modules/FindPackageHandleStandardArgs.cmake:386 (_FPHSA_FAILURE_MESSAGE)
  /home/vudiq/projects/Lakka-LibreELEC/build.Lakka-TinkerBoard.arm-2.1-devel/toolchain/share/cmake-3.6/Modules/FindPythonInterp.cmake:163 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:16 (find_package)


-- Configuring incomplete, errors occurred!
See also "/home/vudiq/projects/Lakka-LibreELEC/build.Lakka-TinkerBoard.arm-2.1-devel/vulkan-loader-1.0.61.1/.armv7ve-libreelec-linux-gnueabi/CMakeFiles/CMakeOutput.log".
Makefile:12: recipe for target 'image' failed
make: *** [image] Error 1
```
Installing Debian package `python3` (with all its dependencies) on the host system solved the issue.

So python3 should either be added to the initial test, that installs other host system packages, or the Lakka-LibreELEC python package should be upgraded, or separate python3 package added as dependency of vulkan-loader.

Ntemis:
i will remove vulkan loader or you can. I see no intention from rockchip to support vulkan on rk3288 now or in the future.

Ntemis:
bye bye

PR #387, opened 2018-02-14
Rockchip: MiQi & TinkerBoard: remove vulkan-loader

ToKe79:
see https://github.com/libretro/Lakka-LibreELEC/issues/378 - Rockchip is not getting vulkan support (anytime soon at least)

### DIFF
--- a/packages/libretro/beetle-psx/package.mk
+++ b/packages/libretro/beetle-psx/package.mk
@@ -35,11 +35,7 @@ PKG_IS_ADDON="no"
 PKG_AUTORECONF="no"
 
 make_target() {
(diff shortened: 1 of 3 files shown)


Ntemis:
Yeap can you try remove --disable wayland from retroarch package too for this soc and test.
Thanks

kivutar:
It is safe to keep `--disable-wayland` in place, in case wayland is in your build dir, it will prevent RA from adding the wayland context.

Ntemis:
Ok then is gtg

Στις 15 Φεβ 2018 21:05, ο χρήστης "Jean-André Santoni" <
notifications@github.com> έγραψε:

PR #391, opened 2018-02-15
Update Retroarch cores

Ntemis:
- Fix Rock64 bootloader issue by revert to 2017.11 stable


Ntemis:
Most cores are tested and work without issues

PR #392, opened 2018-02-15
beetle-psx: remove Rockchip make flags

ToKe79:
Due to
https://github.com/libretro/Lakka-LibreELEC/blob/a2f526048ffbdea0972dfe0794346907df827f2d/packages/libretro/beetle-psx/package.mk#L24
there is no reason having specific make flags for Rockchip


ToKe79:
already done in https://github.com/libretro/Lakka-LibreELEC/pull/387

PR #399, opened 2018-02-19
pkg_all.sh: build/clean a package for all targets

ToKe79:
This script is useful if you want to build package for all targets (e.g. test if version change does not introduce build errors) or to clean a certain package on all targets (e.g. if you want to make sure that the package is rebuilt on next release/nightly build).

Usage examples:
```
./pkg_all.sh build retroarch
./pkg_all.sh clean retroarch
```

### DIFF
--- a/pkg_all.sh
+++ b/pkg_all.sh
@@ -26,7 +26,6 @@ case $1 in
 esac
 
 package=$2
-distro="Lakka"
 declare -i failed=0
 failed_targets=""
 skipped_folders=""
--- a/pkg_all.sh
+++ b/pkg_all.sh
@@ -1,5 +1,4 @@
 #!/bin/bash
-#set -x
 usage() {
 	echo ""
 	echo "$0 <build|clean> <package>"
@@ -37,9 +36,9 @@ for f in $(ls -d build.Lakka-*/) ; do
 	if [ -d $f ] ; then
 		i+=1
 		declare "skip_$i=no"
-		declare "distro_$i=$(echo $f | sed -e 's/build\.\(.*\)-\(.*\)-2.1\(.*\)/\1/')"
-		declare "arch_$i=$(echo $f | sed -e 's/build\.\(.*\)-\(.*\)\.\(.*\)-2.1\(.*\)/\3/')"
-		target=$(echo $f | sed -e 's/build\.\(.*\)-\(.*\)\.\(.*\)-2.1\(.*\)/\2/')
+		declare "distro_$i=$(echo $f | sed -e 's/build\.\(.*\)-\(.*\)-[0-9]\(.*\)/\1/')"
+		declare "arch_$i=$(echo $f | sed -e 's/build\.\(.*\)-\(.*\)\.\(.*\)-[0-9]\(.*\)/\3/')"
+		target=$(echo $f | sed -e 's/build\.\(.*\)-\(.*\)\.\(.*\)-[0-9]\(.*\)/\2/')
 		if [ -d projects/$target ] ; then
 			declare "project_$i=$target"
 			declare "device_$1="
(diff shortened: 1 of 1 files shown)


Ntemis:
What i don't like in this is change is that someone has to keep maintaining it in case of changes but i admit it looks handy to have

ToKe79:
## Maintenance ##
Following variables have to be kept up to date:

`buildprojects` - to define projects that Lakka is being compiled for
`archs_PROJNAME` - to define architectures (`arm`, `aarch64`, `x86_64`, `i386`, etc) for which the PROJNAME project is compiled and is other than `arm` (which is currently default, therefore does not have to be defined for those projects that are compiled for `arm` architecture); multiple architectures can be defined per project (as is in case of `Generic`)
`devices_PROJNAME` - if the project PROJNAME has to be compiled for various devices (as currently is the case for `Rochchip` project); multiple devices can be defined per project
`systems_PROJNAME` - if the project PROJNAME has to be compiled for various systems (as currently is the case for `Allwinner`, `S8X2`, `S805`, `imx6` projects); multiple systems can be defined per project

It is the same as adding a line(s) to `build_all.sh` but with the difference, that you just add the strings into the variables or define new variables.

**Examples:**

New project `Foo` compiled for `aarch64` architecture and devices `Bar` and `Ball`:

- add `Foo` to `buildprojects`
- define new variable `archs_Foo="aarch64"`
- define new variable `devices_Foo="Bar Ball"`

New project `Goo` compiled for `arm` architecture and no specific devices/systems:
- add `Goo` to `buildprojects`

That's all.

In case there are no variables `archs_PROJNAME`, `devices_PROJNAME` and `systems_PROJNAME` the values of variables `archs_default="arm"`, `devices_default=""` and `systems_default=""` respectively are used.

I can add this information directly into the script, so it does not get lost :-)

I use similar (little bit extensive) script to compile the nightly builds: https://github.com/ToKe79/lakka_build_script/blob/master/build_lakka_nightly.sh

kivutar:
You could achieve the same thing in less code by just looping over `ls build.Lakka`.
It would be project agnostic and wouldn't require maintenance.

ToKe79:
`ls build.Lakka*` just gives the build folder name followed by all package build folders - how does that help? This should rebuild / clean just single package.

Maybe you meant `ls -d build.Lakka-*`, but then for Rockchip you don't have the project name in the folder name, only the DEVICE name (if you want to get the PROJECT, ARCH and DEVICE variables from the folder name; you don't need SYSTEM actually, so that could be omitted also in the pkg_all.sh). **Or I misunderstood your point.**

What now I think I could add is a check if a build folder actually exists to skip projects/devices that are not built locally to avoid building the whole toolchain and dependencies for such projects.

ToKe79:
Or as you suggested parsing the `build.Lakka-*` folder names, comparing them with `projects/*` and `projects/*/devices/*` to get what projects/archs/devices should be compiled - this would not require maintenance, but more code, I think at least.
Any suggestions?

ToKe79:
So it is less code :-)
And should require no maintenance and can be used for any distro (Lakka as well as LibreELEC)

kivutar:
Consider submitting this to LE maybe?

ToKe79:
Had to do some minor changes, as I was searching for build.Lakka-*/, will send PR also here.
https://github.com/LibreELEC/LibreELEC.tv/pull/2516

ToKe79:
So - TIL that to clean a package on all targets there is switch --all for clean, e.g. clean --all nano. But you still have to add DISTRO=Lakka in our case, otherwise it cleans in the default distro, which is LibreELEC. There is  no such feature for build script. And also probably does not work with projects with various devices.
https://github.com/LibreELEC/LibreELEC.tv/pull/2516#issuecomment-367595468

PR #400, opened 2018-02-19
libretro-database: update to the latest database

RobLoach:


Ntemis:
Since 1.7.1 is out i see no reason for individual core updates.

RobLoach:
You mean, do all libretro-related packages in one PR?

Ntemis:
If you indent to do 1by1 you would need a week and then 1.7.2 might get out(e.g as a bug fix)

RobLoach:
I intended to update libretro-database, since there are a change we need.

I'm okay with doing them all at once though.

Ntemis:
In my opinion would be better for all of us and we can **all** test at once

Ntemis:
Wondering if you need to re-open this one @RobLoach

PR #405, opened 2018-02-26
add git branch feature

ToKe79:
- libretro_update: add git branch feature
- path: adapt packages metadata

Goal: to be able ls-remote other than HEAD / master branch.
Sameboy* is currently buildable only on the buildbot branch (https://github.com/LIJI32/SameBoy/issues/31).
EDIT: *newer versions


scripts/unpack
```diff
@@ -103,8 +103,9 @@ if [ -d "$SOURCES/$1" -o -d "$PKG_DIR/sources" -o -n "$PKG_GIT_URL" ]; then
   fi
 
   if [ -z "$PKG_URL" -a -n "$PKG_GIT_URL" ]; then
+    [ -n "$PKG_GIT_BRANCH" ] && GIT_CLONE_BRANCH="--branch $PKG_GIT_BRANCH" || GIT_CLONE_BRANCH=""
     rm -rf $BUILD/$PKG_NAME-$PKG_VERSION
-    git clone $PKG_GIT_URL $BUILD/$PKG_NAME-$PKG_VERSION/
+    git clone $GIT_CLONE_BRANCH $PKG_GIT_URL $BUILD/$PKG_NAME-$PKG_VERSION/
```
RobLoach: The branch ends up being ignored because there's a `git checkout $PKG_VERSION` down below.

ToKe79: yes, right, we do not need `--branch` here, as we do not clone single branch, but whole repository, so we should be able to checkout also a commit from different branch. I will test first.

ToKe79: reverted and squashed

PR #406, opened 2018-02-27
desmume: fix/update url

ToKe79:
core was moved (?)

### DIFF
--- a/packages/libretro/desmume/package.mk
+++ b/packages/libretro/desmume/package.mk
@@ -19,11 +19,11 @@
 ################################################################################
 
 PKG_NAME="desmume"
-PKG_VERSION="9464582"
+PKG_VERSION="69658a2"
 PKG_REV="1"
 PKG_ARCH="any"
 PKG_LICENSE="GPLv2"
-PKG_SITE="https://github.com/libretro/desmume"
+PKG_SITE="https://github.com/libretro/desmume2015"
 PKG_GIT_URL="$PKG_SITE"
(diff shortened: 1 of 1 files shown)


gouchi:
Yes it seems there are [three repositories](https://github.com/libretro?utf8=%E2%9C%93&q=desmume&type=&language=) now.

kivutar:
So, are you sure it is the same source code as before @ToKe79 ?

ToKe79:
On the /desmume repo it is not possible to checkout the commit defined in our PKG_VERSION. With this url it is possible.
But with the commits from https://github.com/libretro/desmume2015/commit/69658a23a8fb02c533f1f60c23311c925243d722 the target is `desmume2015_libretro.so`. We have to keep this in mind if we update this core later (update `makeinstall_target()`), or I could also update it to last version now and update it too in this PR (but I have no games to test with).

RobLoach:
Suggest sticking with the unyear'ed one:
https://github.com/libretro/desmume

ToKe79:
@RobLoach but that one is the different one. For a short time the /desmume url was opening/redirecting to /desmume2015

Ntemis:
Imo better follow upstream and ditch 2015

ToKe79:
Just FYI - the new desmume core requires OpenGL, which we do not have for all boards, so we could include it, but only for specific targets/projects/devices.

RobLoach:
s/permanent/temporary

Ntemis:
Fixed, thanks

PR #407, opened 2018-03-01
Update most libretro cores

kivutar:
This updates most of the cores except a few that were not building:

 - citra
 - sameboy

I'm building images with this change, I can tell so far that Generic and RPi2 are building.

I'm going to boot these images on a few computers and see if the cores are still working.

### DIFF
--- a/packages/libretro/retroarch/package.mk
+++ b/packages/libretro/retroarch/package.mk
@@ -19,7 +19,7 @@
 ################################################################################
 
 PKG_NAME="retroarch"
-PKG_VERSION="3a981e0"
+PKG_VERSION="b3d3cbd"
 PKG_REV="9"
 PKG_ARCH="any"
 PKG_LICENSE="GPLv3"
--- a/packages/libretro/retroarch/package.mk
+++ b/packages/libretro/retroarch/package.mk
@@ -19,7 +19,7 @@
 ################################################################################
 
 PKG_NAME="retroarch"
-PKG_VERSION="b3d3cbd"
+PKG_VERSION="29a057a"
 PKG_REV="9"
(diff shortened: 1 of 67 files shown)


ToKe79:
I got build errors with citra (not this PR, but generally) with newer versions. First it complained about cmake, but when I upgraded cmake, it complained about missing header:
```
 #include_next <stdlib.h>
               ^~~~~~~~~~
```
For sameboy we need to switch to "buildbot" branch. So just add `PKG_GIT_BRANCH="buildbot"` to `package.mk` of sameboy and run `./libretro-update.sh -p sameboy`.

packages/libretro/retroarch/package.mk
```diff
@@ -19,7 +19,7 @@
 ################################################################################
 
 PKG_NAME="retroarch"
-PKG_VERSION="3a981e0"
+PKG_VERSION="b3d3cbd"
```
ToKe79: I think we should go with https://github.com/libretro/RetroArch/commit/6fc6bfbb243de4da05a86d1edc3950815a964f1e, which is 1.7.1 release

ToKe79:
My proposed sameboy changes are over at https://github.com/kivutar/Lakka-LibreELEC/pull/2.

kivutar:
There are no real (including ARM) test phase in RA, so checking out the exact 1.7.1 tag doesn't ensure that it is bug free anyway.

ToKe79:
Images for test:
http://nightly.builds.lakka.tv/special_builds/coreupdate01032018/

kivutar:
I updated cores once again. I tested every core on an RPi2. All work except PPSSPP.

PR #410, opened 2018-03-12
allow libretro overlays to be added via samba and the overlayfs 

markwkidd:


Ntemis:
What is the actual benefit of this?

markwkidd:
I think it's the same case as for having a share and overlayfs for `assets`. Just as I am currently using the Samba share for assets to add a few icons that are missing from FlatUI, I would like to work with some overlays that are not built-in (yet).

Specifically I came across new Vectrex overlays that are formatted for RetroArch but not in the official collection (found here: https://retropie.org.uk/forum/topic/2638/vectrex-overlay-artwork)

It would be nice to add those to my Lakka console without having to access via SSH or have to add them github repository.

PR #419, opened 2018-03-18
Merge LibreELEC 8.2.4

kivutar:
This merge should make our images compatibles with the new RPi3 B+

But I see a dangerous change that may break our NOOBS images.

We need to test these.


kivutar:
@Ntemis have you tested NOOBS on this branch?

Ntemis:
I don't see why you have to worry about NOOBS since we are following LE stable branch. I will find it highly unlikely to have any issue with it since is already tested by LE team but even so you can just revert 
https://github.com/libretro/Lakka-LibreELEC/pull/419/commits/861bb0e445f5552f48c4ec4f1ebe03e5b5066899
and keep the RPi 3B+ compatibility that this merge aims for, no?

Ntemis:
@popcornmix shout out for a  RPi 3B+ sample :)

SapphireTactician:
@kivutar Revert? Does this mean the PR is no longer a part of the nightly? I just bought / upgraded to a 3B+ and the nightly as of 3/22 http://nightly.builds.lakka.tv/2018-03-22/RPi2.arm/ still suffers from the same issue described in #422 which was suppose to be resolved by #419 .

Am I the proud owner of a new brick? -_- Would really like an image that I can use on my RPi 3B+ Please, and thank you.

PR #434, opened 2018-03-29
Merge LibreELEC 8.2.4

kivutar:
- [x] Check that Lakka and LE can still dual boot on NOOBS [Confirmed working by SapphireDrew]
 - [x] Fix linux package irq patch preventing Generic to build
 - [x] Address SapphireDrew comment about `/home` being used in place of `/storage` in smb


kivutar:
I reverted the irq commit that was preventing Generic to build.
This PR should be mergeable now.

PR #440, opened 2018-04-02
Update RA to 1.7.2

kivutar:
Most things are working.
Needs deeper testing.


Ntemis:
I don't see the reason you have left all the cores intact.
Will should run a total core update too.

RobLoach:
> I don't see the reason you have left all the cores intact.

Doing one Pull Request for all cores along with RetroArch can get messy. Keeping focused changes, like this one, makes the diff a lot easier to review.

Ntemis:
@RobLoach you mean we should sent core updates one by one?

kivutar:
@Ntemis this is a difficult question. Sending individual core updates would be convenient in some ways, like easy atomic revert, and keeping the PR discussion about a single change. But it would also be a lot of work and time wasted.

If somebody send a bunch of core updates, I will not reject the PR. But my goal is still that we test things *before* merging instead of *after*. To keep master more stable.

This RA update doesn't have core updates because RA takes quite a lot of time to be tested for regressions.

It won't prevent us to do core updates in another PR.

Things like supporting new devices can be done in a later release (and it doesn't have to be 6 month later). I'd like to go back to 1 release per month like before. Maybe 2 releases per month.

RobLoach:
The only bug I've found on RPi is the Menu Toggle...

1. Load a game
2. Toggle the menu
3. Try to un-toggle the menu

Works on desktop using x as the Input Driver and udev as the Joypad Driver. RPi is both udev.

kivutar:
This bug was already there before the update anyway. So this update brings only good things, no regressions so far.

PR #445, opened 2018-04-05
[WIP] Update libretro cores

kivutar:
- [ ] check that all projects compile
 - [ ] regular cores need to be tested
 - [ ] citra needs cmake 3.8
 - [ ] citra fails to build even with cmake 3.8
 - [x] sameboy has been force pushed and its makefile is broken
 - [x] fix ppsspp
 - [ ] fix audio in ppsspp
 - [ ] citra needs to be well tested


packages/libretro/ppsspp/patches/ppsspp-arm-fix.patch
```diff
@@ -0,0 +1,47 @@
...
+ # ARM
+ else ifneq (,$(findstring armv,$(platform)))
+ 	TARGET := $(TARGET_NAME)_libretro.so
+-	TARGET_ARCH = armv7
++	TARGET_ARCH = arm
+ 	fpic := -fPIC
+ 	LDFLAGS += -shared -Wl,--version-script=link.T -Wl,--no-undefined
+-	FFMPEGINCFLAGS += -I$(FFMPEGDIR)/linux/$(TARGET_ARCH)/include
+-	FFMPEGLIBDIR := $(FFMPEGDIR)/linux/$(TARGET_ARCH)/lib
++	FFMPEGINCFLAGS += -I$(FFMPEGDIR)/linux/arm/include
```
gouchi: If you use commit [7b61523](https://github.com/libretro/ppsspp/commit/7b615236497c5af720724b14f283f9a4560e6dcf) $(TARGET_ARCH) will work as it is removing the [override](https://www.gnu.org/software/make/manual/html_node/Override-Directive.html#Override-Directive) otherwise use `override TARGET_ARCH = arm` instead of `TARGET_ARCH = arm`
[
Source](https://github.com/libretro/ppsspp/pull/4#issuecomment-379269642)

kivutar: armv7 is needed for the ffmpeg lines

gouchi: Yes [armv7](https://github.com/hrydgard/ppsspp-ffmpeg/tree/master/linux) for the include/lib dir.

I should have written `override TARGET_ARCH = armv7` ;-)

So we should "force" it to arm dir
```
FFMPEGINCFLAGS += -I$(FFMPEGDIR)/linux/arm/include
FFMPEGLIBDIR := $(FFMPEGDIR)/linux/arm/lib
```

gouchi: I see you put TARGET_ARCH = arm because of [Makefile.common condition](https://github.com/libretro/ppsspp/blob/master/libretro/Makefile.common#L471).

PR #453, opened 2018-04-23
mame2003-plus: Add MAME 2003-Plus package

markwkidd:
Would you consider adding mame2003-plus to the list of available Lakka packages?

There is at least one tester @SapphireTactician who is interested in building Lakka with mame2003-plus.

Is there anything else that needs to be done in order for it to be available?


markwkidd:
@gouchi I have just updated this PR, and I hope that it is complete now. thank you for your help!

markwkidd:
Hi @RobLoach , glad to. I just pushed a commit to fix the merge conflict?

Do you think things look ok now? Anywhere else you can think I need to add in order for mame2003-plus to be available as a package in Lakka?

markwkidd:
@SapphireTactician you should now be able to build Lakka with the mame2003-plus package enabled. I'm not exactly sure how to do that to be honest but we can figure it out together! 

Also you will see a place in the new mame2003-plus package for Lakka where you could type in updated git commit codes to determine which commit of mame2003-plus is part of Lakka. That might be handy given that the project is still evolving so quickly

RobLoach:
> I'm not exactly sure how to do that
```
./scripts/build mame2003-plus
```

markwkidd:
Thanks Rob!

PR #458, opened 2018-04-27
Update Libretro cores to latest commit to be on par with 1.7.2 retroa…

Ntemis:
…rch release

- Removed nestopia upstream patch


RobLoach:
Does this build?

PR #459, opened 2018-04-27
XU[34] Update kernel to 4.14.35 LTS

Ntemis:


RobLoach:
Looks like we've diverged quite a bit more upstream:
https://github.com/LibreELEC/LibreELEC.tv/blob/libreelec-8.2/packages/linux/package.mk

This normal? I don't have the capacity to test this right now unfortunately.

kivutar:
It's normal, there are some changes required in the linux package to support devices that upstream doesn't support.

PR #468, opened 2018-05-04
Update Retroarch & Cores to be on par with V1.7.3 stable

Ntemis:
https://github.com/libretro/RetroArch/releases

### DIFF
--- a/packages/libretro/retroarch/package.mk
+++ b/packages/libretro/retroarch/package.mk
@@ -19,7 +19,7 @@
 ################################################################################
 
 PKG_NAME="retroarch"
-PKG_VERSION="d360265"
+PKG_VERSION="91744d9"
 PKG_REV="11"
 PKG_ARCH="any"
 PKG_LICENSE="GPLv3"
--- a/packages/libretro/4do/package.mk
+++ b/packages/libretro/4do/package.mk
@@ -19,7 +19,7 @@
 ################################################################################
 
 PKG_NAME="4do"
-PKG_VERSION="c8a9402"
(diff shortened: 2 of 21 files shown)


packages/libretro/retroarch/package.mk
```diff
@@ -19,7 +19,7 @@
 ################################################################################
 
 PKG_NAME="retroarch"
-PKG_VERSION="d360265"
+PKG_VERSION="91744d9"
```
RobLoach: May or may not need to force a `--disable-qt`? Probably just sees Qt is not there, and then disables it automatically.

Ntemis: https://github.com/libretro/RetroArch/commit/19f6798947f8a4dcbfcc43aab6d30686dfbf8e54

gouchi:
We should consider to [update ScummVM core](https://github.com/libretro/scummvm/commits/master) as there are some "critical timing bug fix" and new feature "add MIDI support via 'libretro-deps/fluidsynth".

[Source](https://forums.libretro.com/t/scummvm-on-lakka-2-1-is-it-working/16104/7)

ToKe79:
@Ntemis can you update `scummvm` to https://github.com/libretro/scummvm/commit/c65b4171b740d35da19d7ab8d37ee283e9b92aaa? I confirm that all [projects](http://nightly.builds.lakka.tv/special_builds/pr468/) are buildable, but I did not have time to test any image yet. However - this PR does not include the latest merge from LibreELEC 8.2. I will run another build with the latest commits in Lakka-2.1-dev.

PR #471, opened 2018-05-14
[WIP] Nintendo Switch project

natinusala:
I added a Nintendo Switch project to Lakka. It's still a work in progress so don't merge yet, but I'm making this PR to initiate the process and be ready to merge once it's finished. The project is called `Switch`.

## The system

* The kernel is 4.16 from fail0verflow, with patches from Shiny Quagsire to add wired Joy-Cons support
* The system can be booted by using the USB exploit fusée gelée, the method depends on what you want :
  * Any fusée implementation : boot hekate and chainload `coreboot.rom` (an entry should appear as `Lakka`)
  * shofel2 : boot cbfs and sideload `coreboot.rom` over usb

I changed the partitions layout (of the Switch project only) so that the ext4 partition no longer exists. The `/storage` mountpoint is now a bind mount to `/flash/Lakka/storage`.

This allows :

* Better install process : no need to nuke the SD card with etcher or partition manually - this is the greatest improvement as Lakka is not alone on the card
* Better update process : just extract the tar and overwrite the old files
* Reading and writing data using Windows
* Sharing data (ROMs, savefiles) with homebrew RetroArch

## What changed in Lakka

My changes may break other Lakka projects. I **did not** test anything but the Switch port _yet_.

The following packages were **updated** :

* `mesa` : 18.1.4
* `libdrm` : 2.4.92
* `kernel-firmware` : `8fc2d4e`
* `presentproto` : 1.1
* `xorg-server` : 1.19.6
* `libepoxy` : 1.5.2
* `util-macros` : 1.19.2
* `ppsspp` : 1.6, official release from the original repo rather than libretro's fork
    * the arm fix was altered to fix arm64 too - some parts of the patch may be needed to build on other platforms because of the missing `Replay.cpp` rule
* `json-c` : `985c46f`
* `pulseaudio` : 11.99.1
* `desmume` is now `desmume-2015`
* `dolphin` : `535df63`

The following packages were **modified** :

* `retroarch` : added default configuration for the Switch version
* `linux` : added the `kernel-firmware` dependency for the Switch project
* `kernel-firmware` : added a firmware file at project scope
    * A firmware file is missing from the `linux-firmware` repository, however it is present in the Chromium OS branch so I pull it from there
* `xorg-server` : revert a patch only for the Switch project
* `systemd` : backported a patch to fix a build error on some systems

The following packages were **added** :

* `xf86-video-nouveau` (unused)
* `xinput`
* `mergerfs`
* `desmume`

The following scripts were modified :

* `config/graphics` : added nouveau and tegra drivers
* `scripts/mkimage` : added Switch "bootloader" support

## Current state of the port

* The system boots
* Hardware acceleration works at full speed with or without xorg (KMS & DRM)
* Wi-Fi works after a reboot
* Bluetooth works and is stable thanks to new firmware
* SSH & SMB work
* Touch screen works
* No audio through speakers and headphones
* Audio through Bluetooth (AD2P) using PulseAudio
* No graceful shutdown, no software shutdown
* No USB host
* No Switch Dock support
* Only reboots to RCM (tethered boot)
* Wired Joy-Cons support
* RetroArch is patched to :
    * Remove the `Shutdown` option
    * Rename the `Reboot` option to `Reboot (RCM)` (all languages)
    * Add a GPU profile selection menu
    * Add a backlight brightness control menu
* The typical Linux on Switch issues are still present :
    * ~~(Horizon OS) Battery decalibration~~
    * (Horizon OS) RTC decalibration
    * (Linux) Potential temporary LCD damage or burn-in
    * (Linux) No power management / ACPI / fan control
    * ~~(Linux) No SDXC support~~

## What remains to be done before merging

Those are the steps required to make the system _playable_. Once these are done, an experimental image will be released to the public, and this PR will be in a mergeable state. Future improvements will come when the kernel is updated (no ETA).

* ~~Add a GPU clock profile management service and its corresponding option in RetroArch~~
* ~~Fix wired Joy-Cons support and the corresponding input mapping~~
* ~~Fix the rotation issue~~
    * ~~The Switch LCD screen is portrait and not landscape, as a result the image is rotated 90° to the right~~
    * ~~My Lakka port doesn't use xorg so it's not as simple as running `xrandr -o left`~~
    * ~~RetroArch will need to be patched to support full rotation (see [this issue](https://github.com/libretro/RetroArch/issues/6770))~~
* ~~The splash screen also needs to be rotated (trivial)~~
* ~~Finish the default RetroArch configuration~~
    * ~~Remove the online updater as it's not supported on the Switch yet~~
* ~~Add u-boot `mkimage` as a host dependency (used to build the boot script)~~
* Code cleanup
    * ~~Remove the Wi-Fi workaround service (reboots automatically to enable Wi-Fi without having inputs)~~
    * ~~Put the libretro cores back into the project - I removed them to gain compilation time~~
    * ~~Clean the kernel boot command~~
    * Diagnose and fix build issues
    * ~~Clean `platform_init`~~
    * ~~Remove useless drivers~~
* Test every other project to see if I didn't break anything by updating some packages (I might need help on that one)
* ~~Finish the various RetroArch patches~~
    * ~~Wi-Fi should not freeze the UI when connecting to a network~~
    * ~~Add GPU Profile selection menu~~
* ~~Figure out the boot chain : should we package coreboot and u-boot ? If so, how should we handle `tegra_mtc.bin` ?~~
* ~~The storage partition resize script doesn't reboot properly, instead it puts the Switch in a comatose state where it needs to be rebooted again manually (my guess is that it reboots but doesn't turn the LCD screen back on) - this may need a fix~~

## Disabled cores

Some cores were not compiling. I didn't bother searching why, but I suppose some compability issues with ARM64 or misconfiguration.

Therefore, I disabled those cores in the project :
- ~~desmume~~ ARM64 build fixed
- fuse
- mupen64plus
- picodrive
- ~~ppsspp~~ ARM64 build fixed
- puae
- reicast


gouchi:
Hi,

Thank you for the PR.

I think it will be "better" if you use something like as there is several [Tegra SoCs](https://en.wikipedia.org/wiki/Tegra) ?

`PROJECT=Tegra SYSTEM=t210`

Why don't you put specific packages related to the project in projects/Switch/packages directory especially mesa and libdrm ?

natinusala:
There are indeed several Tegra SoCs, however this is specially made for the Switch :
- The kernel is not just a port of Linux for Tegra, it's been remade from scratch for the console
    - The device tree is made for the Switch
    - The defconfig is made for the Switch too
- There is a built-in wired Joy-Con driver in the kernel, only usable on a Switch
    - The inputs are configured for the Joy-Cons by default on RetroArch
- The dummy bootloader is also made for the Switch

I see this project like the Gamegirl project ; it runs on known hardware (Tegra) but it needs additional tweaks and fixes, like the Gamegirl runs on known hardware, the Raspberry Pi. We can do as you want, but I can't guarantee that `PROJECT=Tegra SYSTEM=t210` will work on _all_ `tegra210` devices (it won't). What other devices with a Tegra X1 could we target ? 

About the packages ; can I override a package's version in the project without copying everything else ? I mean, I can copy paste the whole packages to the project but it's not going to be maintainable, especially when there are a lot of files (patches, services, etc...). I don't like duplicating things if it's not necessary.

natinusala:
Current state of the port - see how the display is in portait mode : https://www.youtube.com/watch?v=8hq6v_ScYhw

natinusala:
Updating connman to the latest version broke RetroArch's Wi-Fi menu ; how should I handle that ? Should I only update connman for my project and add a patch for RetroArch, or should I keep the updated package and make a PR to RetroArch to fix the issue / add the patch to every Lakka project ?

kivutar:
The best way would be to understand what changed in connman that breaks the wifi menu. 

If the wifi menu is not listing SSIDs, the reply is somewhere there https://github.com/libretro/RetroArch/blob/master/wifi/drivers/connmanctl.c#L50-L96

If you can diagnose it, you can submit a retroarch patch that fixes the compatibility.

If not, update connman only in your project.

natinusala:
I already know what causes the issue, the output for the commands changed so it breaks the parser in RetroArch. It's very simple to fix.

If I submit a fix in RetroArch directly, you will need to update RetroArch to the latest commit so that it works once the Switch project is merged. Do you want to do this / do you want me to do this ?

Or if you prefer to keep the 1.7.3 stable, merge the 1.7.3 pull request first and then I'll write a patch in a separate PR to Lakka. It's up to you, depending if you want to stay in the stable branch of RetroArch.

natinusala:
Or, another possibility, I can patch connman to revert what broke the parser, if possible. This may be the best solution for Lakka, however RetroArch on its own will not work with an updated connman.

natinusala:
I fixed the connmanctl driver in RetroArch, however there is an issue with the new version : when restarting connman, the Wi-Fi services are not yet available so the following `systemctl connect` command fails. 

The temporary solution I found is to run the connect command in a loop while it doesn't work, to "wait" for the services to become available. This has the side effect of freezing the GUI for a solid 15 seconds when connecting to a network. 

This is not a viable solution ; do you see another way of doing this ?

kivutar:
@natinusala We're about to release a new stable version of Lakka.

After that, we can get your PR merged into the trunk if you are OK to maintain it on the long term.

natinusala:
Alright, sounds good to me. Is there a time window / a deadline for the merge ? Because there are checks to do before merging :

- Fix the Linux header build issue (temp fix is waiting it to fail, removing the linux build folder and try again ¯\_(ツ)_/¯)
- Check that no other project was broken by my changes (or move the changes locally to the project)

And I'd like to wait as a battery driver is on its way. Since the battery calibration issue is refraining a lot of people from using Lakka (and it makes it somewhat dangerous), I feel like this shouldn't be merged until this is fixed (the Linux battery driver is wrong and screws up calibration data in Switch OS, this results in shortened battery life and wrong battery percentages display).

kivutar:
Indeed it would be better to merge it after the battery issues are fixed.

There is no time window, (but make sure to regularly rebase your branch).

natinusala:
The battery bug has been fixed.

natinusala:
The Switch project is now wholesome, as the latest version of the hekate bootloader will allow chainloading Lakka. The switch-bootloader package is no longer a fake bootloader, as it now includes modified versions of u-boot and coreboot.

Two boot methods can now be used :
- (new method) boot hekate and run Lakka through the chainloader (launch the `coreboot.rom` payload)
- (old method) boot cbfs and sideload `coreboot.rom` over USB

Please note that building the project now requires `tegra_mtc.bin`, a proprietary blob, which is not included in the repo.

projects/Switch/packages/ishiiruka/core_info/ishiiruka_libretro.info
```diff
@@ -0,0 +1,12 @@
+display_name = "Nintendo - GameCube / Wii (Ishiiruka)"
```
gouchi: It is not possible to update [core info package](https://github.com/libretro/Lakka-LibreELEC/blob/Lakka-V2.1-dev/packages/libretro/core-info/package.mk) as it is available [upstream](https://github.com/libretro/libretro-super/blob/master/dist/info/ishiiruka_libretro.info).

projects/Switch/packages/ishiiruka/package.mk
```diff
@@ -0,0 +1,54 @@
+################################################################################
```
gouchi: We can try to enable it for other platforms as requested on issue #511.

projects/Switch/packages/ishiiruka/core_info/ishiiruka_libretro.info
natinusala: My original plan was to add the core and let the users decide which one to use (Dolphin or Ishiiruka) on the Switch. However Lakka didn't let me load a .iso with Ishiiruka without a core info, hence this hacky way of adding the core info.

I can make a PR for the core info package but I wanted to wait for the core to be actually finished. I also need to make sure that the core  can load all of Dolphin's file formats.

Just tell me if you want the PR

projects/Switch/packages/ishiiruka/package.mk
natinusala: I have a package ready for Dolphin and Ishiiruka, however the latter is not finished and I feel like we should wait before adding it. 

I wanted to submit a PR to update both cores at once, along with the new core info, but we can add Dolphin only in the meantime. Ishiiruka however is only there for perfs testing purposes.

projects/Switch/packages/ishiiruka/core_info/ishiiruka_libretro.info
gouchi: Sure no problem lets wait a little the core to be polished ;-)

natinusala: Well it's not finished, I don't think all the inputs are there (there is only GC, no Wii), there is no Vulkan / D3D11 support...

natinusala:
Closing this as the Lakka-2.1-dev branch is no longer used.

PR #472, opened 2018-05-15
Gamegirl: fix linux compilation

ToKe79:
After updating kernel for RPi, Gamegirl needs to be updated too. This compiles, however as I do not have the device, I cannot test it.
Test image: http://nightly.builds.lakka.tv/special_builds/GG/


davidperrenoud:
@ToKe79 I could test this. Don't hesitate to ping me if I haven't done it.

davidperrenoud:
The image "Lakka-Gamegirl.arm-2.1-devel-20180515093235-r28187-gbe2bb2580.img" boots successfully, but the background is white because of the default menu color theme.

@kivutar Could you change the default of the Gamegirl to something like Ribbon / Midnight? The shader Bokeh / Midnight also looks very nice on the Gamegirl but is a bit slow.

davidperrenoud:
Yes, the default `xmb_menu_color_theme` should be fixed. Has the order of the colours been changed?

PR #484, opened 2018-06-05
RPi/RPi2: add support for Zeroplus based gamepads

### LINKED ISSUE #482: Installation on Raspberry Zero W a complete disaster
Downloaded from the Lakka site. Installed and all seemed to be OK.
Does not detect my gamepad even though other progs such as RetroPie detect it.
I used a USB keyboard to do WIFI setup.  It scans and finds my wifi signal but when I select it it just hangs.
It seems to be expecting a gamepad input because it shows an on screen keyboard.
I tried downloading from Git but am not sure how to install it because etcher says it is an invalid image.
I tried setting up wifi in config file but that did not help.

ToKe79:
This adds support for Zeroplus based gamepads.
Fixes #482


gouchi:
LGTM ;-)

PR #496, opened 2018-07-01
New libretro core: REminiscence

ToKe79:
# REminiscence
Port of Gregory Montoir's Flashback emulator, running as a libretro core.

Builds test successful for every target.

Runs on Generic.x86_64. Tried to run on S912, but got only black screen. I don't have other devices available to test.

[Images](http://nightly.builds.lakka.tv/special_builds/new-reminiscence/)

### DIFF
--- a/distributions/Lakka/options
+++ b/distributions/Lakka/options
@@ -253,7 +253,7 @@
 
   LAKKA_MIRROR="http://sources.lakka.tv"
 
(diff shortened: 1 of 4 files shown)


ToKe79:
@RobLoach Thank you for the test!
I wonder what is S912 missing that RPi2 has, besides newer kernel.

PR #524, opened 2018-08-17
Fix (and update) sameboy and gearboy packages

natinusala:
* The sameboy repo was deleted and recreated so the original commit ID was not valid anymore
* The gearboy repo was deleted, upstream is used now instead


gouchi:
Thank you.

Build fine for Generic and ARM Tinkerboard device platform.

Ntemis:
LGTM

PR #526, opened 2018-08-23
[WIP] Libretro update

kivutar:
I tested that everything compile properly. I had to fix 2 or 3 packages.

I've not yet tested the images for regressions.

I'll upload a batch.

### DIFF
--- a/distributions/Lakka/options
+++ b/distributions/Lakka/options
@@ -1,7 +1,7 @@
 # Name of the Distro to build (full name, without special characters)
   DISTRONAME="Lakka"
   LIBREELEC_VERSION="devel"
-  OS_VERSION="2.1"
+  OS_VERSION="2.2"
 
 # short project description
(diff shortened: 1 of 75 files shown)


kivutar:
Builds are there http://static.kivutar.me/unstable/

binaryhermit:
The controller mapping for a dualshock 3 controller on a Raspberry Pi 3B+ using the Raspberry Pi 2 build is very messed up....

kivutar:
I'm getting black icons on RPi2, and no controller hotplug. Like if overlayfs was broken. Or config path changed.

Ntemis:
I had the same issue a while back, i remember reporting it

kivutar:
I found the cause of this. RetroArch is not reading /etc/retroarch.cfg anymore. So we need to find another way to set default values for Lakka.

Issue #557, opened 2018-10-11
Fix Sameboy compilation


grillo-del-mal:
#### Which version of Lakka are you using?
* I tested this with the master branch, and the latest 2 tags (2.1 and 2.1.1) 

#### What system hardware are you using?
* Tested this on a couple of x86 computers running Linux and Docker

#### What did you do?
* Tried to build lakka using the instructions on the dockerfile
```
docker build -t lakka .
docker run -it -v $(pwd):/root lakka
```
* Also tested with some enviorment variables
```
docker build -t lakka .
docker run -it -e DISTRO="Lakka" -e PROJECT="RPi" -e ARCH="arm" -v $(pwd):/root lakka
```

#### What did you expect to happen?
* Build process ends normally, I get my binaries ready to install.

#### What happened instead?
* Build process stops when unpacking the sameboy package
```
  UNPACK   sameboy
Cloning into '/root/build.Lakka-RPi.arm-2.2-devel/sameboy-179d3f6'...
remote: Enumerating objects: 4137, done.
remote: Total 4137 (delta 0), reused 0 (delta 0), pack-reused 4137
Receiving objects: 100% (4137/4137), 5.57 MiB | 514.00 KiB/s, done.
Resolving deltas: 100% (2970/2970), done.
Checking connectivity... done.
error: pathspec '179d3f6' did not match any file(s) known to git.
Makefile:27: recipe for target 'image' failed
make: *** [image] Error 1
```

#### Other tests
* Tried to run the process manually using the sameboy repository (guiding myself from what the `script/unpack` script does
  * Got the same error
```
$ clone https://github.com/libretro/SameBoy.git
Cloning into 'SameBoy'...
remote: Enumerating objects: 4137, done.
remote: Total 4137 (delta 0), reused 0 (delta 0), pack-reused 4137
Receiving objects: 100% (4137/4137), 5.57 MiB | 44.00 KiB/s, done.
Resolving deltas: 100% (2970/2970), done.
$ git checkout 179d3f6
fatal: not a git repository (or any parent up to mount point /)
$ git checkout 9ae8304
fatal: not a git repository (or any parent up to mount point /)
$ git checkout 8d69156
fatal: not a git repository (or any parent up to mount point /)
```
* It actually works if you use the head of the buildbot branch
* Also noticed that when searching the commit on the libretro/sameboy repo's github page, it always points to a branchless commit.

#### Posible fixes
[...]

PR #558, opened 2018-10-12
sameboy: Update to archive

RobLoach:
This is a temporary solution, as `master` is currently broken. The PKG_NAME is changed to allow matching the archive extracted folder name.

References: https://github.com/libretro/Lakka-LibreELEC/issues/557


RobLoach:
Correction: https://github.com/libretro/Lakka-LibreELEC/pull/559

PR #559, opened 2018-10-12
sameboy: Update Sameboy

RobLoach:
When we update sameboy, we have to make sure we're referencing the latest commit from `buildbot`.


ToKe79:
Using the provided `libretro_update.sh` does the job - it pulls the latest commit from `PKG_GIT_BRANCH`, if set.

RobLoach:
Very nice. Didn't know about that command.

PR #563, opened 2018-11-02
RK: Upgrade u-boot version to v2018.09 & Add TinkerBoard S support

Ntemis:
- Also switch MiQi default boot device to sdcard *if available*


natinusala:
Those new patches break RK3399 :
- `u-boot-0007-auto-enable-ums-mode-when-TinkerBoard-is-connected-t.patch`
- `u-boot-0008-add-10ms-delay-after-re-enable-EMMC.patch`
- `u-boot-0009-fixed-enter-ums-mode-fail-sometimes.patch`
- `u-boot-0011-added-timeout-when-force-entering-UMS-mode.patch`

Removing them builds fine. I'll report back on booting once the image is finished building.

natinusala:
It boots correctly. Moving the patches from package scope to device scope should do the trick.

PR #568, opened 2018-11-14
desmume, gme, mesen: use git instead of targz

natinusala:
desmume, gme and mesen were using the targz method instead of git. This is now fixed.


packages/libretro/gme/package.mk
```diff
@@ -24,7 +24,8 @@ PKG_REV="1"
 PKG_ARCH="any"
 PKG_LICENSE="GPLv3"
 PKG_SITE="https://github.com/libretro/libretro-gme"
-PKG_URL="https://github.com/libretro/libretro-gme/archive/$PKG_VERSION.tar.gz"
+PKG_SOURCE_DIR="libretro-gme-$PKG_VERSION"
```
RobLoach: PKG_SOURCE_DIR? Is this needed? The other packages don't have it.

natinusala: Indeed, this is a leftover from the targz which is not needed anymore

ToKe79:
just fyi - I pushed .tar.xz files to sources.lakka.tv:
image

PR #575, opened 2018-12-09
Add /tmp/cores overlay to LD_LIBRARY_PATH

dfranusic:
When libretro cores depend on shared libraries it is quite difficult to add anything to Lakka OS since root partition is a **squashfs** read-only partition. My particular issue was the **hatari_libretro.so** which was compiled without IPF support. When I recompiled it and enabled IPF, I also realized that I cannot copy the **ibcapsimage.so.4** dependency to root partition and the only was to run it would be to manually export **LD_LIBRARY_PATH** in **.profile** and run retroarch from shell, which is not acceptable. This pull request adds **/tmp/cores** overlay to **LD_LIBRARY_PATH** to facilitate installation of custom core dependencies.

### DIFF
--- a/packages/libretro/retroarch/scripts/retroarch-config
+++ b/packages/libretro/retroarch/scripts/retroarch-config
@@ -3,3 +3,4 @@
 for i in 5 4 3 2 1 0; do [ -e /dev/snd/pcmC${i}D0p ] && export ALSA_CARD=$i; done
 
 echo "ALSA_CARD=\"$ALSA_CARD\"" > /run/libreelec/retroarch.conf
+echo "LD_LIBRARY_PATH=\"/usr/lib:/tmp/cores\"" >> /run/libreelec/retroarch.conf
--- a/packages/sysutils/busybox/profile.d/98-busybox.conf
+++ b/packages/sysutils/busybox/profile.d/98-busybox.conf
@@ -18,7 +18,7 @@
 
 export HOME="/storage"
 export PATH="/usr/bin:/usr/sbin"
-export LD_LIBRARY_PATH="/usr/lib"
+export LD_LIBRARY_PATH="/usr/lib:/tmp/cores"
 export HOSTNAME=`cat /etc/hostname`
 export PS1="\[\e[1;32m\]\h\[\e[1;32m\]:\[\e[1;34m\]\w \[\e[0m\]\\$ "
 # k0p


Ntemis:
Why dont you fix the hatari core instead?

natinusala:
I agree that for the sake of completion, users should have a way to add shared libraries to the system

RobLoach:
What is ibcapsimage? Don't see reference to a lib like that in https://github.com/libretro/hatari/blob/master/Makefile.libretro .

dfranusic:
@Ntemis, there is nothing wrong with hatari core, it just doesn't support IPF out of the box. I am unaware of the reason why, maybe due to some licensing issues? I have no problem creating a pull request for that if it's in  everyone's interest.

Libretro's version of hatari core uses a predefined `config.h` found here:
- https://github.com/libretro/hatari/blob/master/libretro/include/config.h

This include path is defined in `Makefile.common` (**-I$(LIBRETRO_DIR)/include**):
```
INCFLAGS := -I$(EMU)  -I$(CPU)  -I$(FALCON) -I$(EMU)/includes -I$(DBG) -I$(FLP) \
	-I$(LIBRETRO_DIR) -I$(LIBRETRO_DIR)/libretro-sdk/include 
        -I$(LIBRETRO_DIR)/include -I$(LIBUTILS) -I$(CPU_PREGEN)
```

Default configuration does not enable **IPF** support which depends on external library called **capsimage** and is indicated by the following line in `config.h` (commented out by default):
```
/* Define if you have the capsimage library */
//#define HAVE_CAPSIMAGE 1
```
These are the steps I did to rebuild hatari core with IPF support:
#### 0. Install capsimage library (either build it from source or use a binary package)

#### 1. Enable HAVE_CAPSIMAGE in config.h
```sh
~/hatari $ sed -ie 's/^.*HAVE_CAPSIMAGE.*.*/#define HAVE_CAPSIMAGE 1/' libretro/include/config.h
```

#### 2. Patch Makefile.libretro (Needed for linking phase)
* [patch.txt](https://github.com/libretro/Lakka-LibreELEC/files/2664057/patch.txt)
```sh
~/hatari $ git apply patch patch.txt
```

#### 4. Build hatari core with IPF support
```sh
~/hatari $ make -f Makefile.libretro EXTERNAL_ZLIB=1 HAVE_CAPSIMAGE=1
```

#### 5. Check if **hatari_libretro.so** is properly linked (just in case)
```sh
~/hatari $ ldd hatari_libretro.so
	libm.so.6 => /lib/libm.so.6 (0xfde63000)
	libz.so.1 => /lib/libz.so.1 (0xfde3f000)
	libpthread.so.0 => /lib/libpthread.so.0 (0xfde13000)
	libcapsimage.so.4 => /usr/local/lib/libcapsimage.so.4 (0xfddf7000)
	libc.so.6 => /lib/libc.so.6 (0xfdc96000)
	/lib/ld-linux-armhf.so.3 (0xfffbe000)
	libstdc++.so.6 => /usr/lib/gcc/armv7a-unknown-linux-gnueabihf/7.3.0/libstdc++.so.6 (0xfdace000)
	libgcc_s.so.1 => /usr/lib/gcc/armv7a-unknown-linux-gnueabihf/7.3.0/libgcc_s.so.1 (0xfdab4000)

```

Ntemis:
I see :) thats gtg by me then. @kivutar?

kivutar:
OK thanks

PR #587, opened 2018-12-31
Generic_VK_nvidia: add support for nvidia GPUs using khr_display Vulkan context

mlnlbrt:
For now Lakka supports nvidia GPUs with mesa/nouveau driver only, which is not as good as drivers for AMD or Intel. Lakka is not usable on newer nv GPUs as nouveau support for them is rather barebones.

This PR adds support for nvidia cards using the proprietary driver provided by vendor and `khr_display` Vulkan context available in Retroarch. Basically, this is a DRM/KMS-like approach, where `VK_KHR_display` Vulkan extension implemented in the proprietary driver is used instead. 

Pros:
- no X11 dependencies
- direct-to-display path, which should be good in context of input lag
- ability to fully utilize modern nvidia GPUs or even to bring them to work with Lakka
- brings Vulkan-only features of some cores (beetle-psx, parallel-n64, dolphin, ppsspp...) to Lakka

Cons:
- no OpenGL stuff

I have built the project and tested it on a real hardware (x86_64 PC with a nvidia GPU). Unfortunately, I didn't have a chance to build other projects, but I hope I didn't break anything (changes are rather simple).

GL-only cores which I am aware of are disabled from the build. I have tested the following cores and they seem to work okay:
- all NES cores
- snes9x
- beetle-psx, pcsx_rearmed
- beetle-saturn
- dolphin
- ppsspp
- parallel-n64 

Please review the pull request and maybe run some automatic builds to see if other projects build fine.

I am aware this Vulkan-only approach can be confusing for new users or introduce additional things to maintenance, so feel free to discard the PR. :)


natinusala:
Thanks for the PR !

Do you have the list of all the cores you disabled for this build ? Are performance improvements worth the trade-off of having less cores ?

mlnlbrt:
@natinusala here is the list of disabled cores taken from `projects/Generic_VK_nvidia/options`:

```
  # disable GL-only cores
    LIBRETRO_CORES="${LIBRETRO_CORES//citra/}"
    LIBRETRO_CORES="${LIBRETRO_CORES//craft/}"
    LIBRETRO_CORES="${LIBRETRO_CORES//desmume/}"
    LIBRETRO_CORES="${LIBRETRO_CORES//mupen64plus/}"
    LIBRETRO_CORES="${LIBRETRO_CORES//openlara/}"
    LIBRETRO_CORES="${LIBRETRO_CORES//redream/}"
    LIBRETRO_CORES="${LIBRETRO_CORES//reicast/}"
    LIBRETRO_CORES="${LIBRETRO_CORES//sameboy/}" # does not seem to build
    LIBRETRO_CORES="${LIBRETRO_CORES//yabause/}"
```

So far I have used my custom Lakka build with X11 and nvidia driver enabled and it worked fine. I just wanted to try SSAA Vulkan-only feature of beetle-psx which looks amazing. 

I haven't done any measurements, so it is hard to say. Probably input lag is better comparing to X11+OpenGL-based path.

kivutar:
This is interesting.

In my opinion, we shouldn't merge this in master, but it would be cool to host these builds on our server.

If you're interested in maintaining a build of Lakka, you can join us on Discord https://discord.gg/BNFR4hM and we can give you access to a build server.

natinusala:
If it's a separate project, why shouldn't this be merged ? Unless it breaks other platforms, just provide another image with a list of disabled cores and a list of better working cores.

Ntemis:
This PR doesn't break anything. I don't get why we can't merge it tbh @kivutar

natinusala:
Is the new kernel version `default-nvidia` really necessary ? What do you need it for ?

natinusala:
But it's the same version, same source tree. If you need the patches, you can put them in the project scope instead of creating a entirely new kernel version

mlnlbrt:
@natinusala I would like to use the `default` kernel version, but then I would need to exclude RT patch from being applied somehow. Is it possible?

natinusala:
Oh so you want to _avoid_ patches, and not add some. I'm afraid that there's no clean way to achieve that other than what you already did. It's ok then

mlnlbrt:
@natinusala Yup, that is my goal here.

I have been thinking and maybe it is be possible to use `default` kernel with `Generic_VK_nvidia` project by moving the RT patch to `projects/Generic/patches/linux`. This way, the RT patch would be applied only for `Generic` project. What do you think?

natinusala:
That depends if those RT patches are currently applied to any other platform than Generic

mlnlbrt:
It looks that `default` kernel is used only by `Generic` project, so I think this is the only platform where it is applied. Will try it and update the PR if it will work.

natinusala:
I would love to merge this but it looks like a check is stuck on this PR ?

**Lakka-LibreELEC ** Expected — Waiting for status to be reported

kivutar:
@natinusala OK I've removed this check

PR #614, opened 2019-03-08
Use Estuary as default Theme for Kodi.

chlorophyll-zz:
When building Lakka with Kodi, 
e.g. MEDIACENTER="kodi" in distributions/Lakka/options, 
Use the default Theme Estuary instead of old default Theme name Confluence
Otherwise the build will interrupt.

### DIFF
--- a/distributions/Lakka/options
+++ b/distributions/Lakka/options
@@ -25,13 +25,13 @@
 # Mediacenter to use (kodi / no)
   MEDIACENTER="no"
 
-# Skins to install (Confluence)
+# Skins to install (Estuary)
 # Space separated list is supported,
-# e.g. SKINS="Confluence"
-  SKINS="Confluence"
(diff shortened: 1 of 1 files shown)


natinusala:
I have a question though, why would you want to build Lakka with (an outdated) Kodi ?

chlorophyll-zz:
I want to have a Kodi Addon that starts Retroarch.
I tried the Kodi Addon Gamestarter but that is not available for Mx9 Rk3328. 
So i would create a Mix of Gamestarter scripts with Lakkas Retroarch.

RobLoach:
There's also [retroarch-kodi-addon-LibreELEC](https://github.com/ToKe79/retroarch-kodi-addon-LibreELEC) from @ToKe79.

chlorophyll-zz:
@RobLoach thank you. i close this pr.

PR #632, opened 2019-04-04
reicast: fix segfaults with Generic

mlnlbrt:


toolboc:
@mlnlbrt , the included patch will not work as-is. 

Should there be an associated change included for [packages/libretro/reicast/
package.mk](https://github.com/libretro/Lakka-LibreELEC/blob/d668ac4c5ef820e227ecbcff5ca13ec16293e3ae/packages/libretro/reicast/package.mk) to update `PKG_VERSION`?  

You are referencing changes to line 835 of the Makefile, but the file is only 676 lines long in the currently referenced PKG_VERSION [(2abb34)](https://github.com/libretro/reicast-emulator/blob/514c569eb3d2e415091eeb47aff1b1781e5b13a2/Makefile).

mlnlbrt:
are you sure? https://github.com/libretro/reicast-emulator/blob/2abb342/Makefile

toolboc:
@mlnbrt, you are correct, I must have stumbled into the wrong commit tree when ivestigating the root cause, please disregard.  

However, when I attempt to build reicast using your updated patch I encountered the following when targeting Generic x86_x64:

```
In file included from ./core/libretro-common/include/glsm/glsm.h:30:0,
                 from ./core/rend/gles/gles.h:3,
                 from ./core/rend/gles/CustomTexture.h:25,
                 from core/hw/pvr/drkPvr.cpp:18:
./core/libretro-common/include/glsym/rglgen_headers.h:66:19: fatal error: GL/gl.h: No such file or directory
```

mlnlbrt:
I have just built reicast for Generic / x86_64 target successfully. The error you are getting does not seem to be related to my patch. 

Have you fully built Lakka for Generic / x86_64 at least one time?

toolboc:
@mlnlbrt, it looks like the issue is due to my build environment being based on the included [Dockerfile](https://github.com/libretro/Lakka-LibreELEC/blob/master/Dockerfile) which is missing  dependencies for reicast (mesa-common-dev).

ToKe79:
fixes #662

PR #633, opened 2019-04-05
Generic_VK_nvidia: enhancements and update of packages

mlnlbrt:


natinusala:
I have some questions about the new Vulkan support in Lakka, kinda unrelated to this PR.

Is adding Vulkan support as easy as adding `VULKAN=xxx` in the project options, or are there project specific changes to be made? Same question for the other way around, will Vulkan work with other projects or is the `Generic_VK_nvidia` name hardcoded somewhere in some packages?

Lakka Switch will soon have Vulkan support too (with the new gpu driver), this is why I ask.

mlnlbrt:
As for now, adding `VULKAN=xxx` will make retroarch Vulkan-only, that's all. `Generic_VK_nvidia` is hardcoded in all other places, mainly to turn some cores Vulkan-only, please take a look at this PR:

https://github.com/libretro/Lakka-LibreELEC/commit/9b82c91ebaaf8437fb627a52e1561559a58762e8

It's just a hack to bring Lakka to nvidia users through proprietary driver without affecting rest of the project. The only way to do so without pulling X11 dependencies is by using `VK_KHR_display` retroarch context which is Vulkan-only, so that is why we have Vulkan in this case.

I think it is good idea to replace hardcoded checking for `Generic_VK_nvidia` project in most of these places where it is used, will commit it soon.

Making Lakka both Vulkan and OpenGL/OpenGLES capable in one build will require more effort, I haven't considered it yet.

natinusala:
Okay I see.

> Making Lakka both Vulkan and OpenGL/OpenGLES capable in one build will require more effort, I haven't considered it yet.

Why ?

mlnlbrt:
@natinusala you would need to:
- build retroarch with support for both OpenGL/OpenGLES and Vulkan (easy)
- enable Vulkan support in mesa (probably easy)
- make `glsl-shaders` and `slang-shaders` coexist (probably easy too)
- generate icd json for vulkan-loader

In my opinion, the latest of the above is the most tricky part. 

I don't know how to generate the icd json if vulkan support is provided by mesa, but I guess it just a matter of pointing to a proper .so, which is the same for all supported GPUs (I hope). Sounds easy.

But things get worse if you have more than one vulkan-provider in one build, let's say `nvidia-driver` and `mesa`. In this scenario you would need to figure out at runtime which driver is used and generate icd json on fly. The same goes for OpenGL, but this is done in LibreELEC and works somehow.

Despite all the above, having both Vulkan and OpenGL enabled in retroarch, you would need to switch between `gl` and `vulkan` contexts, depending on your needs at the time, please see https://github.com/libretro/RetroArch/issues/4804.
In my opinion it is in contradiction with plug-and-play spirit of Lakka.

natinusala:
I already have the icd json for my GPU driver so that won't be really hard (and I only need to target one GPU). Despite being a pain in the ass to port to other distributions, I must at least say that their sample rootfs has everything working out of the box.

> Despite all the above, having both Vulkan and OpenGL enabled in retroarch, you would need to switch between gl and vulkan contexts, depending on your needs at the time,

I don't see the issue with changing the video driver and restarting RetroArch, while it's cumbersome it works

mlnlbrt:
kms video_context_driver used by Generic project does not support Vulkan anyway :)

natinusala:
Oh yeah I was not asking for Generic, it's for Lakka Switch with xorg and a copy-paste-all-the-close-sourced-shared-libs driver with Vulkan and GL support

natinusala:
@mlnlbrt did you do anything special to make vulkan work? On Switch, `vulkan-smoketest` works but RetroArch "cannot find instance symbols"...

natinusala:
Tested on Lakka-Switch with Vulkan

mlnlbrt:
@ToKe79 @natinusala thank you guys. It would be great if you merge the PR (`Only those with write access to this repository can merge pull requests.`).

ToKe79:
@mlnlbrt the reason for disabling mupen64plus is only the fact that it supports GL only?

The reason I am asking: I am working on a PR to merge mupen64plus-libretro-nx core (package name mupen64plus_next) and ran into build error with Generic_VK_nvidia project:
```
_next: no package.mk file found
```
The reason was that `LIBRETRO_CORES="${LIBRETRO_CORES//mupen64plus/}"` did not remove only the mupen64plus package form the list, but it also removed the first part of the *mupen64plus*_next package name :-) I also noticed that in package.mk of mupen64plus there is a case condition for "Generic", which is not considered in case of Generic_VK_nvidia and the mupen64plus core fails to build. So I made these changes to the package.mk of mupen64plus:
```
diff --git a/packages/libretro/mupen64plus/package.mk b/packages/libretro/mupen64plus/package.mk
index 4636ea771..af07d89f6 100644
--- a/packages/libretro/mupen64plus/package.mk
+++ b/packages/libretro/mupen64plus/package.mk
@@ -55,18 +55,22 @@ make_target() {
       CPPFLAGS="$CPPFLAGS -DLINUX -DEGL_API_FB"
       make platform=unix GLES=1 FORCE_GLES=1 HAVE_NEON=1 WITH_DYNAREC=arm
       ;;
-    Generic)
-      make
+    Generic*)
+      case $ARCH in
+        x86_64)
+          make WITH_DYNAREC=x86_64
+          ;;
+        i386)
+          make WITH_DYNAREC=x86
+          ;;
+      esac
       ;;
     OdroidC1)
-      make platform=odroid BOARD=ODROID-C1 GLES=1 FORCE_GLES=1 HAVE_NEON=1 WITH_DYNAREC=arm
+      make platform=odroid BOARD=ODROIDC1 GLES=1 FORCE_GLES=1 HAVE_NEON=1 WITH_DYNAREC=arm
       ;;
     OdroidXU3)
       make platform=odroid BOARD=ODROID-XU3 GLES=1 FORCE_GLES=1 HAVE_NEON=1 WITH_DYNAREC=arm
       ;;
-    ROCK960)
-      make platform=unix-gles GLES=1 FORCE_GLES=1 HAVE_NEON=1 WITH_DYNAREC=arm
-      ;;
     *)
       make platform=unix-gles GLES=1 FORCE_GLES=1 HAVE_NEON=1 WITH_DYNAREC=arm
       ;;
```
The other two changes (OdroidC1 and ROCK960) are irrelevant here. Now the core builds. Can you keep the mupen64plus core for Generic_VK_nvidia? It might be not usable with Vulkan (I do not really know and cannot test - I don't have a rig with pascal nvidia card), but it compiles correctly, so it is should be no showstopper including the core in this project as well.

natinusala:
As a side note, Lakka Switch has both gl and Vulkan cohabiting fine. We can change between one and the other by going to Drivers settings.

Why not do the same here? Can't you have gl+kms and vk+khr at same time?

mlnlbrt:
@ToKe79 yes, this is the reason for disabling `mupen64plus` and other cores like `reicast`.
These GL-only cores will not be usable, so that is why I don't want to include them. They will build, because `mesa` has been added as dependency for `nvidia-driver`, but it is only to satisfy build-time requirements of cores like `ppsspp` and `dolphin` IIRC. 

To summarize, I would prefer to add `mupen64plus_next` to the list of disabled cores in `Generic_VK_nvidia` project.

@natinusala there is no KMS for proprietary nvidia driver. `vk_khr_display` context driver is the only way to use binary blobs from nvidia without X11. 

I have been thinking about some "hybrid" version of `vk_khr_display` context, which would enable it to be usable with `gl` drivers (not only vulkan) with help of EGL and offscreen rendering, but I have little OpenGL and no Vulkan knowledge :/

natinusala:
Oh I see. Not surprised to see that the Nvidia driver sucks for desktop too

PR #634, opened 2019-04-07
Set proper LD_LIBRARY_PATH to fix #598

toolboc:
See:
https://forum.libreelec.tv/thread/14975-openssl-compilation-issue/?postID=114573#post114573

PR #645, opened 2019-05-20
openssl: Remove trailing : from the end of LD_LIBRARY_PATH to prevent…

ooPo:
… a segfault during build.


ToKe79:
possible duplicate of #634? different solution thou.
thanks for your contribution!

ToKe79:
the culprit was in `config/path`, I already submitted a PR (#649) with fix including other fixes

Issue #598, opened 2019-02-03
Build segfaults on Debian 9


Zlika:
On Debian 9 / AMD64, there is a segfault when compiling openssl:

```
make[2]: Entering directory '/home/thomas/git/Lakka-LibreELEC/build.Lakka-Generic.x86_64-2.2-devel/openssl-1.0.2o/.x86_64-libreelec-linux-gnu/test'
make[2]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
make[2]: Nothing to be done for 'install'.
make[2]: Leaving directory '/home/thomas/git/Lakka-LibreELEC/build.Lakka-Generic.x86_64-2.2-devel/openssl-1.0.2o/.x86_64-libreelec-linux-gnu/test'
making install in tools...
make[2]: Entering directory '/home/thomas/git/Lakka-LibreELEC/build.Lakka-Generic.x86_64-2.2-devel/openssl-1.0.2o/.x86_64-libreelec-linux-gnu/tools'
make[2]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
make[2]: Leaving directory '/home/thomas/git/Lakka-LibreELEC/build.Lakka-Generic.x86_64-2.2-devel/openssl-1.0.2o/.x86_64-libreelec-linux-gnu/tools'
installing libcrypto.a
installing libssl.a
installing libcrypto.so.1.0.0
installing libssl.so.1.0.0
Segmentation fault
make[1]: *** [Makefile:552: install_sw] Error 139
make[1]: Leaving directory '/home/thomas/git/Lakka-LibreELEC/build.Lakka-Generic.x86_64-2.2-devel/openssl-1.0.2o/.x86_64-libreelec-linux-gnu'
Makefile:26 : la recette pour la cible « release » a échouée
make: *** [release] Erreur 2
```

I found a thread on this compilation bug in the Libreelec forum (https://forum.libreelec.tv/thread/12422-openssl-compilation-issue-for-8-2-5-9/), but I don't know how to fix it.

Regards

gouchi:
Did you try ?

> You have to install libssl1.0-dev instead of libssl-dev on Debian Stretch

Zlika:
Yes. Without success.

bite-your-idols:
I have the same issue using ubuntu 16.04 docker container, just for Generic devices, for RPi it builds fine... any solution?

[...]

PR #649, opened 2019-06-02
Buildfixes

ToKe79:
these are fixes I had to make in order to make the builds work on ubuntu 18.04.02 LTS.
it solves #598, which makes #634 and #645 obsolete.


ToKe79:
Images for testing are here: http://nightly.builds.lakka.tv/special_builds/pr649/

PR #651, opened 2019-06-02
mupen64plus: build flag changes

ToKe79:
+ Generic: added dynarec flags
+ OdroidC1: fixed BOARD flag
+ ROCK960: removed (same as default case)


natinusala:
So for all this time, Generic mupen was without a dynarec?

PR #636, opened 2019-04-08
add missing mkimg via u-boot-tools

toolboc:
*adds missing `mkimg` tool needed by https://github.com/libretro/Lakka-LibreELEC/blob/master/scripts/image#L298

PR #654, opened 2019-06-04
Fix Dockerfile.

halsafar:
Sameboy compilation and image creation is broken in the image generated by the Dockerfile.  Easy fix.

- Add `bsdmainutils` (fix Sameboy compilation).
- Add `u-boot-tools` (fix image creation).


ToKe79:
Thank you, but these changes are already in the PRs #636 and #639.

PR #658, opened 2019-06-06
flycast: renamed from reicast

plaidman:
compiles and runs full speed on switch


packages/libretro/flycast/package.mk
```diff
@@ -18,31 +18,39 @@
...
-    make platform=rpi FORCE_GLES=1 HAVE_OPENMP=0
+    if [ "$OPENGLES_SUPPORT" = "yes" ]; then
+      REICAST_GLES=1
+    else
+      REICAST_GLES=0
+    fi
+
+    make platform=rpi FORCE_GLES=$REICAST_GLES HAVE_OPENMP=0
+  elif [ "$PROJECT" == "Switch" ]; then
+    make platform=arm64 HAVE_OPENMP=0 FORCE_GLES=0 GLES=0 HAVE_OIT=1
```
plaidman: I'm not sure about this tbh. I think I should add a generic one since our switch specific stuff is in another repo

plaidman: this would expose a rebase conflict here but it should be easy to resolve and add the switch-specific config when we get there.

plaidman: or should I leave off aarch64 altogether?

natinusala: No Switch stuff should be present here, take it out

natinusala: I suppose `"$PROJECT" == "aarch64"` is good enough for now, we don't support any aarch64 targets anyway, might as well remove it

natinusala:
I just realized that the whole `$REICAST_GLES` thingy was for the arm Switch port, because the Switch was the first (and only) platform to be armv7 _and_ OpenGL, I had to forcefully disable gles for some cores (reicast and mupen). The makefiles made the (incorrect) assumption that arm == gles, which was false in our case.

I think this can be taken out of the package here, unless there is another platform using arm _and_ opengl that I'm not aware of

natinusala:
I will update core-info and libretro-super once all the new cores are merged

plaidman:
updated core-info and solved a rebase issue with Generic_VK_nvidia options

natinusala:
Can you rebase your branch against the latest changes made to reicast? Make sure to keep the new patch

packages/libretro/core-info/package.mk
```diff
@@ -19,7 +19,7 @@
 ################################################################################
 
 PKG_NAME="core-info"
-PKG_VERSION="44a65bc"
+PKG_VERSION="4a020d4"
```
natinusala: That's wrong, bas copy paste from core package.mk

plaidman: fixed

PR #660, opened 2019-06-07
picodrive: add aarch64 support

plaidman:
compiles and runs full speed on switch

### DIFF
--- a/packages/libretro/picodrive/package.mk
+++ b/packages/libretro/picodrive/package.mk
@@ -21,7 +21,7 @@
 PKG_NAME="picodrive"
 PKG_VERSION="82f7735"
 PKG_REV="1"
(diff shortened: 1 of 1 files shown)


packages/libretro/picodrive/package.mk
```diff
@@ -21,7 +21,7 @@
 PKG_NAME="picodrive"
 PKG_VERSION="82f7735"
 PKG_REV="1"
-PKG_ARCH="arm i386 x86_64"
+PKG_ARCH="arm i386 x86_64 aarch64"
```
natinusala: It should be "any"

plaidman: 👍 all set

PR #661, opened 2019-06-08
genesis-plus-gx: fix segfault at runtime for aarch64

plaidman:
compiles and runs for switch


plaidman:
note that this is quite different than how I did it for the lakka-switch repo. there will be rebase conflicts here, but I like this implementation a lot better.

natinusala:
I like how this is done

PR #667, opened 2019-06-11
dolphin: add aarch64 support and build patch

plaidman:
compiles and runs on switch

### DIFF
--- a/packages/libretro/dolphin/package.mk
+++ b/packages/libretro/dolphin/package.mk
@@ -19,9 +19,9 @@
 ################################################################################
 
 PKG_NAME="dolphin"
-PKG_VERSION="1a7d1bc"
+PKG_VERSION="e8f27d0"
 PKG_REV="1"
-PKG_ARCH="x86_64"
+PKG_ARCH="x86_64 aarch64"
 PKG_LICENSE="GPLv2"
 PKG_SITE="https://github.com/libretro/dolphin"
 PKG_GIT_URL="$PKG_SITE"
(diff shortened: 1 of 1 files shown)


packages/libretro/dolphin/patches/dolphin-01-buildfix.patch
```diff
@@ -0,0 +1,30 @@
...
+    set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--no-undefined")
+-else()
+-   target_link_libraries(dolphin_libretro PRIVATE
+-      core
+-      uicommon
+-      ${LIBS}
+-   )
+ endif()
+    
+-if(LIBRETRO_STATIC)
```
RobLoach: Would there be value in having `LIBRETRO_STATIC` defined by default for the core? Would save having to maintain the patch, and the if() statements around it.

plaidman: I believe specifically having `LIBRETRO_STATIC` defined wouldn't be correct, because that would change the values being passed into `target_link_libraries`, right? But I do like your suggestion of changing the defined values to match what we need to happen, instead of patching the cmake.

Please correct me if I'm wrong about the first thing, I wasn't the original author of the change, so I'm not totally sure if `-static-libgcc -static-libstdc++` is wanted there.

natinusala: @plaidman do you want to make more changes on this before merging the PR?

plaidman: updated to remove patch file. good suggestion @RobLoach

PR #669, opened 2019-06-11
yabasanshiro: new core

plaidman:
compiles and runs on switch. has a problem where the games freeze every few minutes. haven't yet diagnosed.

### DIFF
--- a/distributions/Lakka/options
+++ b/distributions/Lakka/options
@@ -258,7 +258,7 @@
 
   LAKKA_MIRROR="http://sources.lakka.tv"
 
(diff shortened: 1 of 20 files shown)


plaidman:
@natinusala I marked this core as aarch64 only because it hasn't really been tested on other platforms has it? Should we include others? Should we even include this at all for aarch64 since it doesn't really work 100% on switch anyway? When it does run it's full speed but we have the freezing issue that we haven't solved yet.

natinusala:
I would say to add it to all archs, but it will depends on how well it runs / if it runs at all

plaidman:
updated for all archs.

natinusala:
Can you do the same for kronos please?

plaidman:
kronos is already 'any'

natinusala:
I am building images for Generic, RPi2 and ROCKPro64 to test yaba sanshiro and kronos - I will need someone to test Generic for me, and all other platforms if needed (or we can assume that it works on all boards if it works on RPi and ROCKPro64)

PR #670, opened 2019-06-11
kronos: new core

plaidman:
compiles and runs on switch

### DIFF
--- a/distributions/Lakka/options
+++ b/distributions/Lakka/options
@@ -258,7 +258,7 @@
 
   LAKKA_MIRROR="http://sources.lakka.tv"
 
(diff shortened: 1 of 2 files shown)


plaidman:
@natinusala same question here as sanshiro: do we want this to be arch: any or just aarch64? do we even want to include this core at all?

natinusala:
It has come to my ears that kronos only works on OpenGL, which only Generic supports at the moment. The arch should be set to x86 and x86_64 only (yaba sanshiro is fine)

plaidman:
set it to `i386  x86_64` and rebased.

PR #672, opened 2019-06-11
reicast: properly fix segfaults with Generic

### LINKED ISSUE #662: Lakka 2.2.2 Reicast Crash
System: Lakka 2.2.2 - 64Bit on 7 Gen Intel NUC.
What happen: Launch anything with Reicast crash the system.
What should happen: Launch Reicast, game runs. 

All required and optional system/dc files are loaded.

mlnlbrt:
Removing Dockerfile updates which should be no longer needed and restore previous version of reicast patch, hopefully fixes #662

### DIFF
--- a/packages/libretro/reicast/package.mk
+++ b/packages/libretro/reicast/package.mk
@@ -35,14 +35,12 @@ PKG_IS_ADDON="no"
 PKG_AUTORECONF="no"
 
 pre_configure_target() {
-  if [ "$ARCH" != "arm" ]; then
-    CFLAGS="${CFLAGS//-flto/}"
-    CFLAGS="${CFLAGS//-ffat-lto-objects/}"
-    CXXFLAGS="${CXXFLAGS//-flto/}"
-    CXXFLAGS="${CXXFLAGS//-ffat-lto-objects/}"
-    LDFLAGS="${LDFLAGS//-flto/}"
-    LDFLAGS="${LDFLAGS//-ffat-lto-objects/}"
-  fi
+  CFLAGS="${CFLAGS//-flto/}"
+  CFLAGS="${CFLAGS//-ffat-lto-objects/}"
+  CXXFLAGS="${CXXFLAGS//-flto/}"
+  CXXFLAGS="${CXXFLAGS//-ffat-lto-objects/}"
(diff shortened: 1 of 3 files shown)


packages/libretro/reicast/package.mk
```diff
@@ -34,6 +34,15 @@ PKG_LONGDESC="Reicast is a multiplatform Sega Dreamcast emulator"
 PKG_IS_ADDON="no"
 PKG_AUTORECONF="no"
 
+pre_configure_target() {
```
ToKe79: you can replace these lines with `strip_lto`

packages/libretro/reicast/package.mk
```diff
@@ -34,6 +34,15 @@ PKG_LONGDESC="Reicast is a multiplatform Sega Dreamcast emulator"
...
+  CFLAGS="${CFLAGS//-flto/}"
+  CFLAGS="${CFLAGS//-ffat-lto-objects/}"
+  CXXFLAGS="${CXXFLAGS//-flto/}"
+  CXXFLAGS="${CXXFLAGS//-ffat-lto-objects/}"
+  LDFLAGS="${LDFLAGS//-flto/}"
+  LDFLAGS="${LDFLAGS//-ffat-lto-objects/}"
+}
```
ToKe79: my build was failing on missing gnu/stub-32.h so I suggest adding:
```
if [ "$TARGET_ARCH" == "x86_64" ]; then
  CFLAGS="$CFLAGS -D__x86_64__ -D__LP64__
  CXXFLAGS="$CXXFLAGS -D__x86_64__ -D__LP64__"
fi
```

ToKe79: you can disregard this comment. this is a mistake in `config/options`. I was building with simple `make image`. when no PROJECT is set, PROJECT=Generic, when no ARCH is set, TARGET_ARCH=x86_64, but ARCH is left empty. starting the build with `ARCH=x86_64 make image` works.

ToKe79:
Thank you for your PR. I will prepare an image for #662 to test (including suggested change above).

packages/libretro/reicast/package.mk
mlnlbrt: Done. Good catch, thanks!

PR #470, opened 2018-05-08
Added gamecon_gpio_rpi driver for Raspberry Pi

ronj:
I have added (a fork of) the gamecon_gpio_rpi driver to Lakka. This driver enables the usage of gamepads (NES, SNES, N64, PS) directly from Raspberry Pi's GPIO pins.

There is no set-up method included at the moment, so the user needs to manually insert the kernel module with the correct arguments. I.e. in my case, having a SNES controller attached:

    # modprobe gamecon_gpio_rpi map=0,0,0,0,0,1

More information can be found on the RetroPie wiki: https://github.com/RetroPie/RetroPie-Setup/wiki/GPIO-Modules#gamecon_gpio_rpi

I have tested this on my SNES controller with built-in Pi Zero W and it works like a charm.

PR #685, opened 2019-06-16
RPi/RPi2: gamecon_gpio driver

ToKe79:
based on #470
[dnb]

If the original PR does not get rebased and corrected, this is the alternative.


natinusala:
Can you add a README on the package.mk explaining how to enable it?

PR #693, opened 2019-06-22
get-git: new clone when remote changes

ToKe79:
removes original repo folder in case the remote has changed and makes a new clone of the new remote.
[dnb]


natinusala:
This is really good, thanks for that

ToKe79:
The original fix was not working properly + I added some output.

ToKe79:
corrected the commit message

natinusala:
Whoops, wrong account

Why doesn't GitHub allow login from multiple accounts like Google does

PR #697, opened 2019-07-12
pkg_all.sh: drop some targets

ToKe79:
- drop some targets - only difference is uboot, enough to build for one
target from the group
- fix path of build folder - was not working with IGNORE_VERSION=1

[DNB] as no changes to build system/packages

### DIFF
--- a/pkg_all.sh
+++ b/pkg_all.sh
@@ -32,17 +32,7 @@ targets="\
 	RPi|arm||||noobs \
 	RPi2|arm||||noobs \
 	Allwinner|arm||Bananapi||image \
-	Allwinner|arm||Cubieboard2||image \
-	Allwinner|arm||Cubietruck||image \
-	Allwinner|arm||orangepi_2||image \
-	Allwinner|arm||orangepi_lite||image \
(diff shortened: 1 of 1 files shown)


natinusala:
The change in $BOARD will make different build folders, I'm not sure I follow the changes here

ToKe79:
Yes, BOARD makes different build folder, but it is not necessary to test a package build for all BOARDs, as all devices have the same settings, the only difference that BOARD makes it the uboot/cmdline:
https://github.com/libretro/Lakka-LibreELEC/blob/19c6dad5c5da71d09772e5da3c0ef7da5bdb66ab/projects/Rockchip/devices/RK3399/options#L27-L50
So if a build/compilation works for one of those project/device targets, it will work also on the remaining ones. This script is mostly for testing if certain package (e.g. after updating a libretro core, retroarch, etc.) builds for all targets.

ToKe79:
I probably get now where you are headed - for example you have locally only build folder for ROCK960 but not for ROCKPro64, so it would not trigger a build of the package for Rockchip / RK3399. I will add back the BOARD targets, as they have their own build folders.

PR #707, opened 2019-07-16
 Play!: new libretro core - Sony Playstation 2

ToKe79:
I had to modify cmake to use curl from the toolchain (added dependency + configure flag). I expect that the buildbot will fail at first, as the cmake will not be rebuilt. If so, I will clean the build folders completely to avoid any unexpected consequences in other packages.


natinusala:
Is the curl change needed to have cmake download boost when building?

ToKe79:
yes, that is the reason - I saw your hack with the wget, I think this is cleaner. More cleaner would be if it would be possible to enable https for curl in cmake, but I did not find a configure flag to enable that.

natinusala:
Of course this is cleaner

ToKe79:
Hi,
this PR (91b26a0ca9a032b6e12139d6d7fd9a4dec1b0572) did not build correctly for all targets.
Following targets failed (1):
- Generic_VK_nvidia.x86_64

Logs and images for testing can be found here: http://nightly.builds.lakka.tv/PRs/pr_707_91b26a0/

ToKe79:
I killed the rebuild, no need to rebuild -the core is now disabled for Generic_VK_nvidia (no OPENGL/ES support here).

Ntemis:
Am pretty sure vk supports ogl as a backwards compatibility api. You sure is not just a case of missing depends?

natinusala:
I couldn't find any info on VK <-> GL backwards compatibility, are you sure of that?

ToKe79:
sure, if you have the right drivers/headers/middleware

natinusala:
Yeah, but a Vulkan-only driver will not expose a GL interface, which is the issue here

PR #720, opened 2019-07-23
systemd/retroarch overlays: fix systemd ordering

ToKe79:
`systemd-tmpfiles-setup.service` was queued after `local-fs.target`, which
also mounted `/tmp` and in order to mount overlays, we need the workdir to
be created (by `systemd-tmpfiles-setup.service`), but `tmp.mount` (on which
the overlay `tmp-*.mount` units depend) is queued before `local-fs.target`.
this resulted into endless cycle and systemd broke the cycle by removing
one of the units, which resulted e.g. the worker dirs not to be ready or
skipping the `tmp-*.mount` units for overlays.

now `systemd-tmpfiles-setup.service` is queued after `tmp.mount` and
`var.mount` (instead of `local-fs.target`) and the queue is correct, the
overlays mount correctly.

### DIFF
--- a/packages/sysutils/systemd/package.mk
+++ b/packages/sysutils/systemd/package.mk
@@ -187,7 +187,12 @@ post_makeinstall_target() {
   sed -e "s,^.*HandleLidSwitch=.*$,HandleLidSwitch=ignore,g" -i $INSTALL/etc/systemd/logind.conf
   sed -e "s,^.*HandlePowerKey=.*$,HandlePowerKey=ignore,g" -i $INSTALL/etc/systemd/logind.conf
 
-  # replace systemd-machine-id-setup with ours
+  # fix ordering cycle for overlay mounts
+  if [ "$DISTRO" = "Lakka" ]; then
+    sed -e "s/local-fs\.target/tmp.mount var.mount/" -i $INSTALL/usr/lib/systemd/system/systemd-tmpfiles-setup.service
+  fi
+
+# replace systemd-machine-id-setup with ours
   safe_remove $INSTALL/usr/lib/systemd/systemd-machine-id-commit
   safe_remove $INSTALL/usr/lib/systemd/system/systemd-machine-id-commit.service
   safe_remove $INSTALL/usr/lib/systemd/system/*.target.wants/systemd-machine-id-commit.service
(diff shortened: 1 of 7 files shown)


packages/sysutils/systemd/package.mk
```diff
@@ -187,7 +187,12 @@ post_makeinstall_target() {
   sed -e "s,^.*HandleLidSwitch=.*$,HandleLidSwitch=ignore,g" -i $INSTALL/etc/systemd/logind.conf
   sed -e "s,^.*HandlePowerKey=.*$,HandlePowerKey=ignore,g" -i $INSTALL/etc/systemd/logind.conf
 
-  # replace systemd-machine-id-setup with ours
+  # fix ordering cycle for overlay mounts
+  if [ "$DISTRO" = "Lakka" ]; then
```
natinusala: Do we really care about breaking the LibreELEC distro? I think this condition could be dropped

ToKe79: It is easier for resolving conflicts (when merging libreelec) if we know why it was added and that it is for Lakka. I did other changes to upstream files like this too, i.e.:
https://github.com/libretro/Lakka-LibreELEC/blob/4a39bb14245dec24746bdc0223798587a74a3db4/packages/graphics/libdrm/package.mk#L56-L57
or
https://github.com/libretro/Lakka-LibreELEC/blob/596ea200359c0896f5c531aa7348c790ca5adefa/packages/graphics/mesa/package.mk#L61-L64

natinusala: I feel like a comment does the same, like `### LAKKA ###` or whatever thing we can easily grep

ToKe79: Yes, a comment would do, but I think we should care about not breaking LE, if it is possible. A one condition extra does not cost extra maintenance. It is extra "code", even if it would be just one line, or one line in a condition. Maybe in future we will need do more changes to the systemd units and this condition would come handy - "oh, the changes can go here" ;-) Again, comment would do the same job, but would break LE and that is not what my OCD allows me to do :-D

natinusala: Fair enough, I find it a little heavy to add the condition to all the packages we edit but if it's how you want to do I don't mind

PR #724, opened 2019-07-24
RPi4: switch back to OPENGLES

ToKe79:


packages/graphics/bcm2835-driver/package.mk
```diff
@@ -19,6 +19,29 @@ else
   PKG_FLOAT="hardfp"
 fi
 
+post_unpack() {
+  # do not build GLES stuff when not using as GLES driver
+  if [ "${OPENGLES}" != "bcm2835-driver" ]; then
```
natinusala: If we want to be precise it should check for both $OPENGL and $OPENGLES, we never know in case the future versions of mesa expose a higher version of OpenGL that we can use

ToKe79: If we would use OPENGL="something", then OPENGLES="" so the condition will be again true, so the OPENGLES stuff will be stripped too. it will be false (and skip the stripping) only in case OPENGLES="bcm2835-driver". Or you mean for a case OPENGL="bcm2835-driver"? Like

     if [ "${OPENGLES}" != "bcm2835-driver" -o "${OPENGL}" != "bcm2835-driver" ]; then

natinusala: With a `-a` - we only want to remove it if gles isn't the bcm driver and gl isn't the bcm driver

Even so there's a very low chance that this driver gets updated anymore, should we care?

ToKe79: We should if by any chance it will get updated.

PR #729, opened 2019-07-28
add dragonboard to uboot_helper

ToKe79:
these commits are cherry-picked from LibreELEC.tv/master in order to allow image build of Qualcomm Dragonboard.


natinusala:
Should this not be squashed then?

ToKe79:
we can squash these.

PR #742, opened 2019-08-06
gearsystem : add gearsystem libretro core

wwark:
Hi,
just tested on generic x64.

### DIFF
--- a/distributions/Lakka/options
+++ b/distributions/Lakka/options
@@ -383,7 +383,6 @@
 "/tmp/cores/genesis_plus_gx_libretro.so;"\
 "/tmp/cores/genesis_plus_gx_libretro.so;"\
 "/tmp/cores/genesis_plus_gx_libretro.so;"\
-"/tmp/cores/gearsystem_libretro.so;"\
 "/tmp/cores/picodrive_libretro.so;"\
 "/tmp/cores/genesis_plus_gx_libretro.so;"\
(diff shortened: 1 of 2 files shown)


ToKe79:
Hi,
the build of this PR (71ed5895f752f2e233263817e27e1eb1a74b4ff0) finished successfully for all 45 targets.
Images for testing and logs are here: http://nightly.builds.lakka.tv/PRs/pr_742_71ed589/

distributions/Lakka/options
```diff
@@ -383,6 +383,7 @@
 "/tmp/cores/genesis_plus_gx_libretro.so;"\
 "/tmp/cores/genesis_plus_gx_libretro.so;"\
 "/tmp/cores/genesis_plus_gx_libretro.so;"\
+"/tmp/cores/gearsystem_libretro.so;"\
```
ToKe79: this line has to be removed, otherwise the array/list is not correct. in this list (`RA_PLAYLIST_CORES`) only one core per system/playlist (`RA_PLAYLIST_NAMES`) is allowed. when you added the gearsystem core, you have shifted all cores below this line to different playlist/system. you have modified the default core for "Sega - PICO", if that was your intention, then remove the line below, but I would suggest keeping the default cores as they are - so just remove this line.

wwark:
Done ! Sorry I think I have to have something there.
Thank you for the explanation.

PR #752, opened 2019-08-16
Lakka: change online updater URL for nightlies

ToKe79:
if users try to upgrade nightly build, they are offered to upgrade to
latest stable release, which is a downgrade. a patch corrects the URL to
the nightlies server. this patch is applied only when the build is
started with LAKKA_NIGHTLY=yes, so it has no impact in other scenarios.

The nightlies build script will take care of populating the `.updater` folder with next nightly build.

### DIFF
--- a/packages/libretro/retroarch/package.mk
+++ b/packages/libretro/retroarch/package.mk
@@ -92,6 +92,17 @@ PKG_CONFIGURE_OPTS_TARGET="--disable-vg \
                            --enable-zlib \
                            --enable-freetype"
 
+post_patch() {
+  # patch URL to nightly builds instead of stable builds if building nightly
+  if [ "$LAKKA_NIGHTLY" = yes ]; then
+    i="${PKG_DIR}/patches/nightly/retroarch-lakka_nightly_url.patch"
+    if [ -f "$i" ]; then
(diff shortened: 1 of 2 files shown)


ToKe79:
Hi,
the build of this PR (56b45dfb7fca22af6b00259ebf09da2810caf806) finished successfully for all 45 targets.
Images for testing and logs are here: http://nightly.builds.lakka.tv/PRs/pr_752_56b45df/

natinusala:
This looks like a sane change to me - could you do even better and upstream the patch to Retroarch? Set the `LAKKA_NIGHTLY` define in the package.mk and use that in the code with an ifdef

ToKe79:
as soon as https://github.com/libretro/RetroArch/pull/9318 is merged and we update RA to the following stable version, we can revert this change (remove the patch, `post_patch()`, adapt `make_target()`. until then I propose to keep this for the nightlies.

natinusala:
Fair enough, we'll remember soon enough when we update RA the patch will fail

natinusala:
It looks like the `.index` file isn't rebuilt automatically

http://nightly.builds.lakka.tv/.updater/.index returns a 404 error

PR #766, opened 2019-09-02
tic80: update for disabled demo carts

RobLoach:
The change was merged forwards, so we can now update with that change in place. We'll continue to use our own remote though, in case there are any further modifications we want to rapidly push upstream:
https://github.com/libretro/TIC-80

### DIFF
--- a/packages/libretro/tic80/package.mk
+++ b/packages/libretro/tic80/package.mk
@@ -19,7 +19,7 @@
 ################################################################################
 
 PKG_NAME="tic80"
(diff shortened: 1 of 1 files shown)


ToKe79:
Hi,
the build of this PR (a453f0bfa21bbaf4fcf208f2943575d32850fabb) finished successfully for all 45 targets.
Images for testing and logs are here: http://nightly.builds.lakka.tv/PRs/pr_766_a453f0b/

PR #770, opened 2019-09-06
get (git): fix fail when branch changes

ToKe79:
this fixes failure when we switch a branch of a package and when the last used branch does not exist on the remote. also adds more output during the various git stages/actions.

### DIFF
--- a/scripts/get
+++ b/scripts/get
@@ -106,33 +106,60 @@ if [ -n "$PKG_GIT_URL" -a -z "$PKG_URL" ]; then
       # no tarball on mirror - clone and create tarball
       CUR_PWD="$(pwd)"
       GIT_REPO_FOLDER="$SOURCES/$1/repo"
+      REPO_FRESH=0
 
       # remove stale folders if they exist
       rm -rf $SOURCES/$1/$PKG_NAME-*/
 
       if [ ! -d "$GIT_REPO_FOLDER" ]; then
-	printf "%${BUILD_INDENT}c ${boldcyan}GIT CLONE${endcolor}    $1\n" ' '>&$SILENT_OUT
+        printf "%${BUILD_INDENT}c ${boldcyan}GIT CLONE${endcolor}    $1\n" ' '>&$SILENT_OUT
         git clone $PKG_GIT_URL $GIT_REPO_FOLDER
+        cd $GIT_REPO_FOLDER
+        REPO_FRESH=1
+      else
+        cd $GIT_REPO_FOLDER
+        REPO_FOLDER_ORIGIN=$(git remote get-url origin || echo ERROR)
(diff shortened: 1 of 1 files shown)


natinusala:
Can you describe roughly how that works?

ToKe79:
`REPO_FRESH` is set to `0` = will be set to `1` if we make a fresh clone (to avoid duplicate code for checking out a branch).

if there is no `repo` folder in `sources/.../`, we make a clone of the repo + set `REPO_FRESH=1`. otherwise we `cd` into the folder and check, if the remote URL changed. if it changed, we make a fresh clone (print a warning, remove the old repo folder, clone) + set `REPO_FRESH=1` (note to self - fix whitespace on line 118 - spaces instead of tab).

if we have a fresh clone (`REPO_FRESH -eq 1`) and we have a branch specified in `PKG_GIT_BRANCH`, we try to checkout that branch, afterwards we check, if we were successful; if not, we bail out.

if we did not make a fresh clone (`else`), we clean any the repo from any manual changes that might be there (`git clean -fdx` + `git checkout -- .`), if we should be using a branch (`PKG_GIT_BRANCH`), we check if we we are on that branch. if we are not on that branch, we `git fetch` new refs from remote and try to checkout that branch (note to self - again checking `-n $PKG_GIT_BRANCH` is redundant - I will remove it there) and in case we are not successful, we bail out.

so now we are sure, we have correct remote, correct branch - we do `git pull` and rewind the history to the desired commit (`git reset --hard ...`). the rest is as before, only added printf to verbose the stage of the GET script.

I will fix those two above (whitespace, redundant `if`) and squash/force push the pr.

natinusala:
Okay so the new behaviour when the branch is invalid is failure with a nice message, instead of the generic git message we have now?

ToKe79:
there will be the git message as well + "nice" message `ERROR: could not checkout branch 'blah' - does not exist!`

ToKe79:
hmm... I have to fix that - if git checkout $PKG_GIT_BRANCH fails, it will not continue... i have to add `|| ...` and put the message there...

ToKe79:
okay.. one more point - we should probably use `git checkout -b $PKG_GIT_BRANCH origin/$PKG_GIT_BRANCH`, as in specific case of mupen64plus_next `git checkout GLideN64` will not fail, but will be caught by the test afterwards.
so if we use `git checkout $PKG_GIT_BRANCH origin/$PKG_GIT_BRANCH` we can avoid the test after the checkout, but we have to handle the failure using `|| { print some nice output; exit 1; }`.

natinusala:
Do what you think is the best, I trust your git-fu better than mine

ToKe79:
ah, I forgot to squash...

PR #777, opened 2019-09-10
Bump version to 2.3

natinusala:
I left `LIBREELEC_VERSION` to `devel`, I will change it to `official` when building locally (along with `OFFICIAL=yes`), tell me if it's not the way to do it (should I set it to `2.3` too instead?)

### DIFF
--- a/distributions/Lakka/options
+++ b/distributions/Lakka/options
@@ -1,7 +1,7 @@
 # Name of the Distro to build (full name, without special characters)
   DISTRONAME="Lakka"
   LIBREELEC_VERSION="devel"
-  OS_VERSION="2.2"
+  OS_VERSION="2.3"
 
 # short project description
   DESCRIPTION="The DIY retro emulation console"


ToKe79:
Hi,
this PR (b1b5e25ae52ca5b36776e1b58235260016831c10) did not build correctly for any target - all (46) builds failed.
Logs can be found here: http://nightly.builds.lakka.tv/PRs/pr_777_000078/

natinusala:
_You saw nothing_

https://github.com/libretro/Lakka-LibreELEC/commit/fc7b79c2cf29b046bdabe74c0cd120587cd874e0

ToKe79:
The correct way is to set `LIBREELEC_VERSION="2.3"`, at least, according to `scripts/image` - it first checks, if `LIBREELEC_VERSION` is set to `"devel"`, then it puts the commit hash/revision number to `LIBREELEC_VERSION`, which we do not want. Then it checks whether `CUSTOM_VERSION` env is defined/set (not our case). And then sets `TARGET_VERSION="$LIBREELEC_ARCH-$LIBRELEC_VERSION"` - we want here e.g. `Generic.x86_64-2.3`, as `IMAGE_NAME="$DISTRONAME-$TARGET_VERSION"`, we should set `LIBREELEC_VERSION="2.3"`.

Yes, the change to `LIBREELEC_VERSION` should be only local and you should build from "dirty" tree (without committing the changes) + adding `OFFICIAL=yes` env to the `make`.

natinusala:
Okay, that's what I thought

So, the commit of that PR is correct right?

ToKe79:
yes, but I would prefer the merge commit of this PR to be tagged as 2.3 (or anything after, to have at least OS_VERSION=2.3 in the tree).

natinusala:
Yeah, I wanted to merge then create the tag, is there an issue with that? Did I miss something?

PR #778, opened 2019-09-10
Lakka: LIBRETRO_CORES style change

ToKe79:
main change is that each core in the LIBRETRO_CORES list is defined on
separate line, which makes the diff with the changes to the list
(new core/removed core) more clear.

another change is, that when removing a core from the list for specific
platform, spaces before and after the core name are added. this ensures
that no partials are removed from the list and the list does not contain
non-existent packages (e.g. to keep flycast_wince, but to remove
flycast, without leading/trailing space the list would contain
non-existent package named _wince).

and last change is in the virtual package for Rockchip - removed
beetle-psx from the PKG_DEPENDS_TARGET, as it is already in
LIBRETRO_CORES.


ToKe79:
Hi,
the build of this PR (df8ffabd423b914d7e4409d6e73b68e93349687f) finished successfully for all 46 targets.
Images for testing and logs are here: http://nightly.builds.lakka.tv/PRs/pr_778_000079/

ToKe79:
nope, beetle-psx only in:
```
build.Lakka-Generic.i386/beetle-psx-fc06bbe/:
build.Lakka-Generic_VK_nvidia.x86_64/beetle-psx-fc06bbe/:
build.Lakka-Generic.x86_64/beetle-psx-fc06bbe/
```

natinusala:
Then I think @Ntemis didn't enable it correctly in his original commit https://github.com/libretro/Lakka-LibreELEC/commit/2b9e41dba7067823395a59feb334d5842170acc4

ToKe79:
the current `package.mk` is quite different, so ... even if `arm` arch would be enabled, it would fall under simple `make`, as neither `VULKAN` nor `OPENGL` is defined (or defined as `"no"`) for any Rockchip platform.

ToKe79:
but at the time of the commit, `PKG_ARCH` was `any`, but it was changed by me (#343) to `x86_64 i386`, probably by mistake.

ToKe79:
Hi,
the build of this PR (a30bcb9580f570cf03f16a45366731af9ca5b8d2) finished successfully for all 46 targets.
Images for testing and logs are here: http://nightly.builds.lakka.tv/PRs/pr_778_000084/

PR #783, opened 2019-09-12
beetle-psx: enable on any arch + GLES

ToKe79:


natinusala:
Hm, seeing that on Switch _with Vulkan_ it runs worse than PCSX ReARMed, I'm not sure this change is welcome for low-end platforms

PR #786, opened 2019-09-14
build_all.sh: start in dashboard mode

ToKe79:
modification of the build script to start in dashboard mode. instead of
showing the output of the individual package compilation a dashboard is
shown showing the overall progress of the current build.

[DNB] do not build by buildbot (no changes to the packages/projects/etc.)

### DIFF
--- a/build_all.sh
+++ b/build_all.sh
@@ -48,7 +48,21 @@ tc="THREADCOUNT=${bt}"
 [ "${DASHBOARD_MODE}" = "yes" ] && v="" || v="-v"
 
 # remove any existing images / release files
-rm -rf target/
+check=$(ls target/ | wc -l)
+if [ ${check} -gt 0 ]
+then
+	echo -en "\nWARNING!!!\nThere are ${check} item(s) in target/, really remove them?\n(press y to confirm, anything else to abort) "
+	read -rn1 keypress
+	if [ "${keypress}" = "y" -o "${keypress}" = "Y" ]
+	then
+		echo -en "\nCleaning up target/ ... "
+		rm -rf target/
+		echo "done."
+	else
+		echo -e "\nAborted!"
+		exit
+	fi
+fi
 
 # list of targets/platforms in structure PROJECT|DEVICE|ARCH|make_rule
 
--- a/build_all.sh
+++ b/build_all.sh
@@ -107,10 +107,13 @@ do
 	target_name=${device:-$project}.${arch}
 
 	# initialize return value of the non-dashboard job
-	non_db_ret=0
+	ret_nondb=0
 
 	# initialize result variable for dashboard job
-	failed_dashboard=0
+	ret_dashboard=0
+
+	# initialize result variable for release files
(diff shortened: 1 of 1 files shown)


ToKe79:
all done here

ToKe79:
now it is complete :)

ToKe79:
OK, I think it is in finished state now, not touching it anymore, so I wont break it.

ToKe79:
I have one more idea - currently, after all packages are built, it just sits on a dashboard with all build threads showing idle (and the dashboard does not update anymore), while the release/image files are created (but nothing is shown to the user - it might look like it is doing nothing). I will add messages that inform the user about that (as soon as all build threads are IDLE).

ToKe79:
any additional ideas / comments / **needs fixing** or can we merge this?

natinusala:
Feature-wise it looks great to me, even more than what I expected

I trust you that it works, however I have questions regarding `UBOOT_SYSTEM`

It's the equivalent of our current `BOARD`, it's for different boards with the same SoC and bootloader where the only difference is the device tree

For instance rk3399 uses it: https://github.com/LibreELEC/LibreELEC.tv/blob/libreelec-9.2/projects/Rockchip/devices/RK3399/README.md

And I even think that the rk3399 device will refuse to build if you don't specify `UBOOT_SYSTEM`

I think it should be added to `build_all.sh`, and all the corresponding targets too (so all RK3399 variants, all RK3328 etc

ToKe79:
You do not need to add `UBOOT_SYSTEM`. If you do not set `UBOOT_SYSTEM`, all possible targets will be build, that are defined in the script `uboot_helper`:
https://github.com/libretro/Lakka-LibreELEC/blob/dbe0b5663abb77128fe087178b732200724e9301/scripts/uboot_helper#L18-L212

It returns the config file and dtb that should be used for the device. So when giving e.g. "PROJECT=Rockchip DEVICE=RK3399`, it will build 5 images (see also here: http://nightly.builds.lakka.tv/Lakka-LE9.2/latest/RK3399.arm/), each suffixed with the `UBOOT_SYSTEM`.

You can also add `UBOOT_SYSTEM` and then it will build only that specific target.

natinusala:
Okay then, do we want build_all to list all the targets comprehensively? To make it clear exactly what we build

ToKe79:
We do not have to, even we should not use `UBOOT_SYSTEM` in the `targets` variable.

The logic is, that in case `UBOOT_SYSTEM` is defined, the temporary files, from which the images are built, get wiped and then with the next job they have to be re-created.

In case `UBOOT_SYSTEM` is empty the install script uses a for loop and re-uses the temporary files in `RELEASE_DIR` to create images for all `DEVICE_BOARDS` and wipes temporary files after.

Here is the relevant part from `scripts/image`:
https://github.com/libretro/Lakka-LibreELEC/blob/dbe0b5663abb77128fe087178b732200724e9301/scripts/image#L308-L337

PR #816, opened 2019-10-03
core-info: add mupen64plus back

natinusala:


ToKe79:
Hi,
the build of this PR (a33668e3eb6bc4395d61b2165b7a942262aa4504) finished successfully for all 46 targets.
Images for testing and logs are here: http://nightly.builds.lakka.tv/PRs/pr_816_000106/

natinusala:
No it doesn't, this is why I added it like so

If you have a less-hacky variant I'll take it but this is good enough I think

ToKe79:
It's the best we can do.

PR #819, opened 2019-10-05
Fixed issue: Could NOT find ZLIB

tbocek:
Fixed: Could NOT find ZLIB (missing: ZLIB_LIBRARY ZLIB_INCLUDE_DIR) when building with docker

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,13 +36,14 @@ RUN apt-get update \
 		sed \
 		tar \
 		texinfo \
-		u-boot-tools \		
+		u-boot-tools \
 		unzip \
 		wget \
 		xfonts-utils \
 		xsltproc \
 		xz-utils \
 		zip \
+		zlib1g-dev \
 	&& rm -rf /var/lib/apt/lists/*
 
 ENV HOME /root


distributions/Lakka/options
```diff
@@ -280,7 +280,6 @@
 " beetle-wswan "\
 " bluemsx "\
 " bsnes "\
-" bsnes_hd "\
 " bsnes-mercury "\
```
ToKe79: please, put `bsnes_hd` back

packages/network/bluez/package.mk
```diff
@@ -1,31 +1,17 @@
...
 PKG_LICENSE="GPL"
 PKG_SITE="http://www.bluez.org/"
-PKG_URL="https://mirrors.edge.kernel.org/pub/linux/bluetooth/$PKG_NAME-$PKG_VERSION.tar.gz"
+PKG_URL="https://www.kernel.org/pub/linux/bluetooth/${PKG_NAME}-${PKG_VERSION}.tar.xz"
 PKG_DEPENDS_TARGET="toolchain dbus glib readline systemd"
-PKG_SECTION="network"
-PKG_SHORTDESC="bluez: Bluetooth Tools and System Daemons for Linux."
 PKG_LONGDESC="Bluetooth Tools and System Daemons for Linux."
+PKG_TOOLCHAIN="autotools"
+PKG_BUILD_FLAGS="+lto"
```
ToKe79: these two have no effect on LibreELEC 8.2 build system, but no problem. if there would be `-lto`, `strip_lto` would be needed.

ToKe79:
seems like backported from LibreELEC master.

tbocek:
my bad, only wanted to do a pull request with the missing ZLIB

tbocek:
@ToKe79 Yes, bluez is a backport as I'm having issues with bluetooth, but it did not resolve it, so no need to update. Also bsnes_hd is not in http://sources.lakka.tv/, so I had to remove it to build it.

ToKe79:
which package was that? because there is zlib, probably that package, which failed, is only missing "zlib:host" in the dependencies.

tbocek:
Its the following issue with cmake:

```
-- Using system-installed ZLIB
-- Could NOT find ZLIB (missing: ZLIB_LIBRARY ZLIB_INCLUDE_DIR) 
CMake Error at CMakeLists.txt:366 (message):
  CMAKE_USE_SYSTEM_ZLIB is ON but a zlib is not found!
Call Stack (most recent call first):
  CMakeLists.txt:662 (CMAKE_BUILD_UTILITIES)


-- Configuring incomplete, errors occurred!
See also "/root/build.Lakka-OdroidXU3.arm-2.3-devel/cmake-3.10.0/.x86_64-linux-gnu/CMakeFiles/CMakeOutput.log".
See also "/root/build.Lakka-OdroidXU3.arm-2.3-devel/cmake-3.10.0/.x86_64-linux-gnu/CMakeFiles/CMakeError.log".
---------------------------------------------
Error when bootstrapping CMake:
Problem while running initial CMake
---------------------------------------------
Makefile:29: recipe for target 'image' failed
make: *** [image] Error 11
```

ToKe79:
can you add `zlib:host` to `PKG_DEPENDS_HOST` of `packages/devel/cmake/package.mk`? it should then install zlib + headers in the toolchain.

ToKe79:
but if it is `cmake:host` that is failing to build, then it will not help, as zlib has `cmake:host` as dependency 🤦‍♂

tbocek:
I tried zlib:host and it ends up in a weird state. I was expecting an error/warning, but nothing happens. Seems just stuck after applying patches.

ToKe79:
probably due to the circular dependency `cmake:host` depends on `zlib:host` and `zlib:host` depends on `cmake:host`... nevertheless, on my buildserver I removed `zlib1g-dev`, ran `./scripts/uninstall cmake:host` and tried to build `cmake:host` agan (with `zlib:host` already being built) and it failed on missing zlib headers as well.

Ntemis:
can we close it?

tbocek:
I haven't tried running a fresh docker build recently, but I believe zlib is still required. However, you can close it, and I will reopen if I still have this issue

tbocek:
I'm closing it for now, there is no Dockerfile in master anymore

PR #824, opened 2019-10-09
Backports from master to LE9.2 branch

ToKe79:


ToKe79:
Hi,
this PR (9b6b48e194822d50b9b622c22dcdb8a1ade5b831) did not build correctly for all targets.
Following targets failed (1):
- Gamegirl.arm

Logs and images for testing can be found here: http://nightly.builds.lakka.tv/PRs/pr_824_000111/

natinusala:
While having the new core info as a patch is _cleaner_ in a way, I think that for our sanity it's better to have them in the `files` directory. It's easier to add new core info files this way, and it's clear where they come from

ToKe79:
Hi,
this PR (26dc3d6237ae125acfe622d22571bf5c2d03ad58) did not build correctly for all targets.
Following targets failed (1):
- Gamegirl.arm

Logs and images for testing can be found here: http://nightly.builds.lakka.tv/PRs/pr_824_000112/

natinusala:
I didn't mean it like that, I think this is actually the worst way of doing that because if we decide to change the location of core info files we also need to change the folder here :sweat_smile: 

But rather make a `files` folder in the package and have the package.mk copy all the files there

Actually I think we could even do it build-script wide, if a `files` folder is present in the package it can copy the files to .install_pkg (between build and install for the package)

ToKe79:
but then you also need full paths under `files` folder, if you want to apply it globally. for now I'll redo it as it is in master (`files` folder in `core-info` package + modify `package.mk`), and maybe later we can come up with something... more general :-)

natinusala:
Uh... yeah that's right. Let's make it on a per package basis for now then

ToKe79:
Hi,
this PR (af4920f6b782f34d4b869f3530f40b2877a8376d) did not build correctly for all targets.
Following targets failed (1):
- Gamegirl.arm

Logs and images for testing can be found here: http://nightly.builds.lakka.tv/PRs/pr_824_000113/

PR #902, opened 2020-02-21
OdroidGoAdvance: Set full screen resolution

shantigilbert:
This avoids crashes when running under Weston


shantigilbert:
This does not apply anymore, closing

PR #915, opened 2020-03-21
Update Theodore core

Zlika:
This new version of the Theodore core adds emulation of new Thomson computers (TO7 and TO7/70).


natinusala:
Thanks for the PR!

PR #933, opened 2020-04-15
Fix rpi host build

dmrlawson:
I was experimenting building this on a Raspberry Pi 4 and hit two issues due to assuming the host was 64bit. These changes fixed it.

### DIFF
--- a/packages/security/nss/package.mk
+++ b/packages/security/nss/package.mk
@@ -20,8 +20,19 @@ make_host() {
   make clean || true
   rm -rf $PKG_BUILD/dist
 
+  local HOST_USE_32_OR_64=""
+  case $(getconf LONG_BIT) in
+    64 )
+      HOST_USE_32_OR_64="USE_64=1";
+    ;;
+    32 )
+      HOST_USE_32_OR_64="USE_X32=1";
+    ;;
+  esac
+
+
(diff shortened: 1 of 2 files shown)


Ntemis:
building where?

Ntemis:
@natinusala what you think about this?

natinusala:
1. Why do you build Lakka on a Raspberry Pi
2. How many days does it take?
3. Do we want to support 32bit hosts? Why would we want to?
4. If that PR does not break 64bit then I don't see why we wouldn't merge it

dmrlawson:
> Why do you build Lakka on a Raspberry Pi

I'm experimenting to see whether I can get any better performance by compiling with -march:native

> How many days does it take?

It took less than a day!

> Do we want to support 32bit hosts? Why would we want to?

I don't know the answer to this, however given that it only takes these two fairly trivial changes to make it work I don't see why not. Whether it's something that you want to support in the future is questionable, but I believe the Raspberry Pi is a popular target and being able to build on it it potentially useful.

natinusala:
I won't support it because there is simply no point. However if those changes don't break 64bit build I am not against merging that PR

Ntemis:
It requires a nifty big sd card size , plenty of time to kill and good cooling.
I see no reason to this what so ever.

PR #941, opened 2020-04-25
Use rpi4 platform for mupen64plus next

dmrlawson:
I noticed it's setting `platform=unix`, possibly from before rpi4 had it's own config.

This works for me on RPi 4


Ntemis:
You assumed right

natinusala:
Thanks for the PR!

PR #955, opened 2020-05-24
Added support for building Lakka in a docker container

tbocek:
Updated and adapted from https://github.com/libretro/Lakka-LibreELEC/tree/Lakka-2.3.2

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+build.Lakka*
+sources
+target
+.ccache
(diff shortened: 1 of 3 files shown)


natinusala:
Thanks for the PR! We don't use Docker but it will certainly be useful for those who do.

autofire372:
Neat; this should make it (technically) possible to build Lakka on Windows!

PR #957, opened 2020-05-26
OdroidGoAdvance: Add support for OGA v1.1

shantigilbert:
This adds support for the OGA v1.1 (Black Edition)

* Bumps retroarch-joypad-autoconfig  to 7d6c1d4 (includes Oga v1.1 game-pad mapping)
* Bumps OGA Linux to 26e571d
* Bumps OGA u-boot to e7b255b
* Modified boot.ini to include dtbs for each revision
* Slight modification to the v1.0 game-pad mapping as I feel this gives more options (fast foward, load/save state, etc) please test to make sure its up to your liking. 

I did not include a rev 1.1 game-pad config as one already exists on the retroarch-joypad-autoconfig package and I am not sure how that is handled, but I think one should be created with all the hotkeys.

### DIFF
--- a/packages/libretro/retroarch-joypad-autoconfig/package.mk
+++ b/packages/libretro/retroarch-joypad-autoconfig/package.mk
@@ -19,7 +19,7 @@
 ################################################################################
 
 PKG_NAME="retroarch-joypad-autoconfig"
-PKG_VERSION="66cb07f"
+PKG_VERSION="7d6c1d4"
 PKG_REV="1"
 PKG_ARCH="any"
 PKG_LICENSE="GPL"
--- a/packages/linux/package.mk
+++ b/packages/linux/package.mk
@@ -23,8 +23,8 @@ case "$LINUX" in
     PKG_SOURCE_NAME="linux-$LINUX-$PKG_VERSION.tar.gz"
     ;;
   odroidgoA-4.4)
-    PKG_VERSION="857cdf6a2ff525fbbe7cf03cf88ef3688a83af1f"
-    PKG_SHA256="0332fe90d0c10a16ad9ea6de93dccc4ca0f15d2da088623f08096a74da9a8977"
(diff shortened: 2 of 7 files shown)


natinusala:
Why do you update retroarch-joypad-autoconfig AND change the file in sources?

My goal was to remove that .cfg and use the one upstream but I never got the chance to merge the two configs

shantigilbert:
I can remove it, I just updates it to have more functionality, but if one upstream is going to exist then that also works for me, I will remove it and push the commit

natinusala:
Can you rather PR it to https://github.com/libretro/retroarch-joypad-autoconfig and then update the repo version here ?

Be careful when merging the two files because I see vala already made a PR recently

shantigilbert:
> Can you rather PR it to https://github.com/libretro/retroarch-joypad-autoconfig and then update the repo version here ?
> 
> Be careful when merging the two files because I see vala already made a PR recently

I would rather someone else do this with the gamepad config my rev 1.0 is all kind of screwed up and I am not sure I can trust it.

natinusala:
Isn't it the same config between the two?

shantigilbert:
> Isn't it the same config between the two?

Unfortunately no, v1.1 has 2 extra buttons and the mapping changes :(

natinusala:
Okay that makes sense. I take that you send the odroidgo2_joypad_v11 autoconfig to upstream and updated the repo commit accordingly here?

shantigilbert:
They already exists (not by me)

https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/udev/odroidgo2_joypad.cfg
https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/udev/odroidgo2_joypad_v11.cfg

natinusala:
Okay then. I might have a look at both of them and make sure all hotkeys used by Lakka are bound (volume, pause/resume, menu, ff...)

Ntemis:
@natinusala any reasons this is stalled?

PR #971, opened 2020-06-22
Bump everything to 1.8.9 Retroarch release head

Ntemis:
- Tinkerboard: Build tested with unbuildable cores disabled

PR #973, opened 2020-06-27
Uncomment DISTRO_SRC and DISTRO_MIRROR so that sources can be downloaded

dmrlawson:
Not sure whether DISTRO_MIRROR is necessary but RPi targets can't be built without DISTRO_SRC at the moment.

Perhaps this change was intentional and the package files need fixing instead? I couldn't find any information about it.


dmrlawson:
Fixed in #971

PR #976, opened 2020-07-10
[RPi] Remove unneeded patch

SethBarberee:
Had a build error with RPi from this. Turns out this was fixed in libretro-common so let's remove the patch.


Ntemis:
Thank you

PR #978, opened 2020-07-12
duckstation : Add duckstation libretro core

gouchi:


ToKe79:
Hi,
this PR (04f29069d720f0b7e5c7477618e8a00467cfdacf) did not build correctly for any target - all (16) builds failed.
Logs can be found here: http://nightly.builds.lakka.tv/PRs/pr_978_000158/

Ntemis:
@gouchi good work, please correct the "cmake-make" as make is called by default and it compiles correct just out of pure luck

gouchi:
@Ntemis Ty. I have just specified `cmake-make` to use `make` instead of `ninja` which is used if we specified only `cmake`.

PR #982, opened 2020-07-15
Added retrodream core to replace redream

Yqnn:
The open source branch of redream core is now called retrodream: https://github.com/libretro/retrodream

Libretro-super is already up to date:
https://github.com/libretro/libretro-super/blob/master/dist/info/retrodream_libretro.info

### DIFF
--- a/distributions/Lakka/options
+++ b/distributions/Lakka/options
@@ -244,6 +244,7 @@ DISTRO_SRC="http://sources.libreelec.tv/$LIBREELEC_VERSION"
 " prosystem "\
 " puae "\
 " px68k "\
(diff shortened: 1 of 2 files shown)


natinusala:
Thanks for the PR! I assume the core works on most platforms?

Yqnn:
I tested it on X86_64.

It should work on the same platforms as Flycast, as they are very similar (but some games works only on one or the other).

twinaphex:
Don't open a PR for this yet, we aren't even sure if the persons that we made this repo for like Rinne will continue working on it. Let's just close this for now.

PR #986, opened 2020-07-19
libgpg-error: add i386 support

ToKe79:
[dnb]
fixes build failure for Generic.i386 target


Ntemis:
This was removed because it was causing build to fail, are you sure we can restore this?

ToKe79:
huh? I do not see any reason for it to cause any failure on other targets/archs, and it works for OK for 32 bit.

Ntemis:
alas it was. ok re adding it and we see as we go

PR #987, opened 2020-07-19
u-boot-tools: fix the sha-hash

ToKe79:
[dnb]
fixes build issue with targets that use u-boot-tools:host (e.g. Amlogic)


ToKe79:
solved by fdb57561c3797ebac83ce688ce276b8a4a47d21a

Ntemis:
Yes we are only using mkimage from this one so lets keep it on same version as upstream le-9.2
Thank you for the pr anw :)

PR #1010, opened 2020-08-16
Add PineH64 Model B uboot configuration

vilhelmgray:


vilhelmgray:
This requires a rebase because the PineH64 Model B device tree files was only added in Linux kernel version 5.6.

PR #1053, opened 2020-10-07
rpi4-aarch64: add parallel-n64

basilean:
I tried it and it works much better than mupen64.


Ntemis:
I don't see why not :)
Thank you for the update

basilean:
Thank you guys!
I got my rpi4 on monday and started to play with first aarch64 image, you made heavy work there. =)

PR #1057, opened 2020-10-08
Switch to more updated uae4arm upstream repo

Chips-fr:
Game compatibility is better with this version.


Ntemis:
Thank you for the pr

PR #1062, opened 2020-10-14
Aarch64 add yabause

basilean:


Ntemis:
you are doing it again buddy

basilean:
@Ntemis Sorry man... I'm not sure how they even are showing there... I'm gonna take it out... the important ones about n64 and psx are already there.... this one works really slow... and mame2015 makes a huge 120mb binary... I will figure them out another day.
Regarding vulkan... looks like there is a long way there... I'm gonna give it a try once in a while hopping one day it will work.

PR #1076, opened 2020-10-26
NXP/iMX6: add fix Wandboard device tree for i2c bus recovery

by-gnome:
In Wandboard scl/sda are stuck low during transfer sometimes, this can happen while reading EDID from TV and lead to no sound. Add scl/sda gpios defitions to device tree to apply i2c bus recovery.


Ntemis:
Happy that you found the issue, good work

PR #1080, opened 2020-11-02
linux: Copy from dts/broadcom/*.dtb when PROJECT=RPi and ARCH=aarch64

dmrlawson:
Previously tried to look at PLATFORM which isn't set.

I decided to check PROJECT as well because we're specifically copying broadcom files here, and we might want aarch64 for other RPis.

Also swapped the order as this is the 'special' case and it makes sense (to me at least) to have the default 'else' case as the previous behaviour.

### DIFF
--- a/packages/linux/package.mk
+++ b/packages/linux/package.mk
@@ -279,10 +279,10 @@ makeinstall_target() {
 
     # install platform dtbs, but remove upstream kernel dtbs (i.e. without downstream
     # drivers and decent USB support) as these are not required by LibreELEC
-    if [ "$PLATFORM" != "RPi4" ] && [ "$ARCH" != "aarch64" ]; then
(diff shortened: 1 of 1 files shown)


ToKe79:
Hi,
this PR (954a4180324530ca8f0d8888adb888f95dae89db) did not build correctly for all targets.
Following targets failed (16):
- A64.arm
- H3.arm
- H6.arm
- AMLG12.arm
- AMLGX.arm
- OdroidXU3.arm
- MiQi.arm
- OdroidGoAdvance.arm
- RK3328.arm
- RK3399.arm
- TinkerBoard.arm
- GPICase.arm
- RPi.arm
- RPi2.arm
- RPi4.aarch64
- Dragonboard.arm

Logs and images for testing can be found here: http://nightly.builds.lakka.tv/PRs/pr_1080_000181/

PR #1081, opened 2020-11-03
Downgrade binutils to fix build errors

davidgfnet:
Cannot rollback to previous version since it does not build with gcc 9/10
However this version seems to work with gcc 10 *and* also produce valid binutils
that are not affected by the ABI change (see https://lkml.org/lkml/2020/1/30/1005)

### DIFF
--- a/packages/devel/binutils/package.mk
+++ b/packages/devel/binutils/package.mk
@@ -3,8 +3,8 @@
 # Copyright (C) 2018-present Team LibreELEC (https://libreelec.tv)
 
 PKG_NAME="binutils"
-PKG_VERSION="2.35"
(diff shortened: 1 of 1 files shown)


davidgfnet:
Sorry that I broke the builds for most plaftorms! Did only check x64, RPI4 and OdroidGoA (the ones I can actually test myself)

Ntemis:
Dont worry i was ready to revert your changes but happy you pushed the fix

davidgfnet:
I was looking for a patch that worked with 2.35 but it requires patching all kernels and each version has a slightly different patch due to tools/ changes over the last kernel versions.
If you know a way to not build tools/ with the kernel that could save build time and similar build errors in the future. However I do not know whether we use the tools in the final image.

PR #1094, opened 2020-11-10
fix build issue in Debian 11

zhangn1985:
this PR aims to fix all target build issue in Debian 11.

current pass:
- [x] A64
- [x] H3
- [x] H6
- [x] AMLG12
- [x] AMLGX
- [x] x86_64
- [x] i386
- [x] MiQi
- [x] RK3328
- [x] RK3399
- [x] TinkerBoard
- [x] Gamegirl
- [x] GPICase
- [x] RPi
- [x] RPi2
- [x] RPi4
- [x] Dragonboard
- [x] iMX6
- [ ] build_all


Ntemis:
Looks good, thanks for your contibutions

distributions/Lakka/options
```diff
@@ -512,3 +512,4 @@ DISTRO_SRC="http://sources.libreelec.tv/$LIBREELEC_VERSION"
       LIBRETRO_CORES+=" mame2000  fbalpha2012 "
     fi
   fi
+LIBRETRO_CORES="${LIBRETRO_CORES// chailove /}"
```
by-gnome, line 515: I'm afraid this will disable` chailove` for all targets.

zhangn1985: yes, many target already disable `chailove`

packages/libretro/melonds/package.mk
```diff
@@ -34,6 +34,10 @@ PKG_TOOLCHAIN="make"
 PKG_IS_ADDON="no"
 PKG_AUTORECONF="no"
 
+if [ "$TARGET_ARCH" = "x86_64" ]; then
```
zhangn1985, line 37: I guess I should remove this check, right?

distributions/Lakka/options
Ntemis: tbh am ok with this

RobLoach: Me too! :+1:

packages/libretro/melonds/package.mk
Ntemis: sure melonds is x86_64 only now

distributions/Lakka/options
zhangn1985: build issue resolved, reenable it.

packages/libretro/melonds/package.mk
zhangn1985: @Ntemis there is a bug, melonds is x86_64 only, but it will add to build plan, thus mesa will add too, and build failed for the projects do not use mesa.

zhangn1985: I don' know how to fix it, thus keep the check.

zhangn1985:
build_all is still in process, but I think it's ready for review.

PR #1099, opened 2020-11-12
filter out melonds

zhangn1985:
melonds will be added to build plan for all targets, so does mesa.

and the targets which don't use mesa will not config mesa correctly.
thus build failed.

Signed-off-by: Zhang Ning <832666+zhangn1985@users.noreply.github.com>

### DIFF
--- a/distributions/Lakka/options
+++ b/distributions/Lakka/options
@@ -224,7 +224,6 @@ DISTRO_SRC="http://sources.libreelec.tv/$LIBREELEC_VERSION"
 " mame2003-plus "\
 " mame2010 "\
 " mame2015 "\
-" melonds "\
 " meowpc98 "\
 " mesen "\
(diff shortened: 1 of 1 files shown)


distributions/Lakka/options
```diff
@@ -512,3 +510,7 @@ DISTRO_SRC="http://sources.libreelec.tv/$LIBREELEC_VERSION"
       LIBRETRO_CORES+=" mame2000  fbalpha2012 "
     fi
   fi
+# add melonds to x86_64
+  if [ "$ARCH" = "x86_64" ]; then
+    LIBRETRO_CORES+=" menlonds "
```
Ntemis, line 515: you have a typo here

zhangn1985: thanks.

PR #1104, opened 2020-11-17
Add kernel install locks to prevent race conditions

davidgfnet:
This uses the locking mechanism provided by the build system to grab a lock
(namely firmware-install) every time a firmware needs to be installed.

This is similar to the mechanism that python uses (exec_thread_safe)
but less restrictive since we do not grab a global lock.


ToKe79:
Hi,
this PR (012220c5eed367e6b6cc36d5390ce4fefa11c3fc) did not build correctly for all targets.
Following targets failed (4):
- H3.arm
- H6.arm
- RK3399.arm
- RPi2.arm

Logs and images for testing can be found here: http://nightly.builds.lakka.tv/PRs/pr_1104_000188/

davidgfnet:
Seems it did not work as I expected :D 
RK3399 is still failing due to fw install, this time during linux:install which is odd. I was not aware that the kernel would install any firmware!

PR #1105, opened 2020-11-18
Adding RG351P support, an Odroid Go Advance clone

davidgfnet:
This console requires a separate DTS and a USB controller config.
U-Boot cannot reliably distinguish the two consoles since the HW ID
for it is the same as OGA v1.1 (looking into the schematics did not
yeld any GPIO differences that could be used for that).

I plan to dig a bit more into u-boot, perhaps enabling USB and
looking for the USB controller could be a way to differentiate them.
In the meantime this just adds another OGA flavor, which adds little
overhead to the build system.

Another TODO is rumble support, mapped to PWM1, which is not trivial
to support in RetroArch.


glebm:
Here is how batocera.linux distinguishes OGA from RG351P on boot:

https://github.com/batocera-linux/batocera.linux/blob/5ebbe17ffcb7a91a44c014e9cf170f394157e31a/board/batocera/rockchip/odroidgoa/boot/boot.ini#L12-L21

glebm:
PWM initialization:

https://github.com/batocera-linux/batocera.linux/blob/5ebbe17ffcb7a91a44c014e9cf170f394157e31a/board/batocera/rockchip/odroidgoa/fsoverlay/etc/init.d/S01pwm#L3-L10

Ntemis:
Good to know. Now we wait for those you have it to test it and sent a pr

davidgfnet:
If you look at the schematic that pin is mapped to the headphone detection signal. I'd imagine (cannot test since my OGA is rev1.0) this only works if there's not headphones plugged in during boot. Perhaps you can test it :)
I have both schematics (OGA and RG351) and I could not find a single GPIO difference to do such a hack in the uboot script. I'm happy to share the schematic with you if you wanna give it a try.

davidgfnet:
> PWM initialization:
> 
> https://github.com/batocera-linux/batocera.linux/blob/5ebbe17ffcb7a91a44c014e9cf170f394157e31a/board/batocera/rockchip/odroidgoa/fsoverlay/etc/init.d/S01pwm#L3-L10

Does that mean that in batocera the console vibrates until you are past kernel boot? I prefer the uboot "hack" since it stops rumbling much sooner :)

iokzero:
@davidgfnet No Batocera don't vibrates for that long, thought it is hard to tell whether or not it is past kernel boot. But definitely sooner the merrier.

PR #1116, opened 2020-12-01
Add MT7601 support for OGA/RG351P

davidgfnet:
Some consoles feature built-in MT7601U cards (also users can use a USB
dongle version of it).


Ntemis:
I dont see why not! Thanks

PR #1119, opened 2020-12-02
[Rpi2/3] linux 5.4

SethBarberee:
Compiles on my system and worked for my RPi3. Played a little pokemon on mgba. No regressions so far from my testing.

Would love someone to check the kernel config to make sure everything looks good but aside from th


projects/RPi/devices/RPi2/patches/bluez/bluez.patch
```diff
@@ -0,0 +1,58 @@
+The 32-bit SIOCGSTAMP has been deprecated. Use the deprecated name
```
by-gnome, line 1: I guess this patch can be moved to `/packages/network/bluez/patches` and then should be removed from projects `/patches/bluez`.

Ntemis: yes it can

PR #1120, opened 2020-12-03
clean up Rpi2/4 patches for linux5.4 and bluez

SethBarberee:
Cleaned up the patches from the RPi2/4 linux 5.4 as suggested in #1119.


projects/RPi/devices/RPi4/patches/bluez/bluez.patch
```diff
@@ -1,58 +0,0 @@
-The 32-bit SIOCGSTAMP has been deprecated. Use the deprecated name
```
by-gnome, line 1: Should also be removed:
projects/Generic/patches/bluez/bluez.patch
projects/NXP/devices/iMX6/patches/bluez/bluez-01-tools-fix-siocgstamp.patch
projects/OdroidXU3/patches/bluez/bluez-32bit.patch

SethBarberee: Removed all three, thx

by-gnome (approved):

PR #1122, opened 2020-12-05
update libretro cores

SethBarberee:
Cores left to update
* beetle-psx (updated to commit before static c++ linking)

Rest of these cores have been compiled on RPi2 and X86_64 target and built fine.


Ntemis:
lgtm ty

packages/libretro/easyrpg/package.mk
```diff
@@ -37,6 +37,7 @@ PKG_IS_ADDON="no"
 PKG_AUTORECONF="no"
 
 PKG_CMAKE_OPTS_TARGET="-DPLAYER_TARGET_PLATFORM=libretro \
+		       -DPLAYER_WITH_FREETYPE=OFF \
```
gouchi, line 40: Just indentation ;)

packages/libretro/pcsx_rearmed/patches/arm_thread_rendering.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/Makefile.libretro b/Makefile.libretro
+index 87b35f7..20b29d5 100644
+--- a/Makefile.libretro
++++ b/Makefile.libretro
+@@ -56,6 +56,7 @@ ifeq ($(ARCH), arm)
+ ifneq ($(findstring __thumb__,$(shell $(CC) -E -dD $(CFLAGS) include/arm_features.h)),)
+ 	# must disable thumb as recompiler can't handle it
+ 	CFLAGS += -marm
++	THREAD_RENDERING = 0
```
SethBarberee, line 9: A Band-Aid on arm but makes it compile.. unless someone finds something better

by-gnome:
It will be more convenient if you use commit messages like:
easyrpg: fix indent
pcsx_rearmed: disable threaded renderer on arm
ppsspp: update ignore sdl patch
Thank you.

packages/libretro/pcsx_rearmed/patches/arm_thread_rendering.patch
SethBarberee: Patch isn't needed anymore

SethBarberee:
* beetle-psx has static c++ linking issues bc of last 2 commits so  I set version to right before it. 
* bsnes2014 and bsnes-mercury have profiles now. Guidance on default choices would be appreciated. My bsnes commits are mainly copying the default profile that it selects.

Otherwise PR is good to go.

PR #1154, opened 2021-01-02
Fix mupen64plus-nx runtime error when using libmali driver

### LINKED ISSUE #1153: Mupen64plus-next crashes with segfault using libmali driver
Since October 2019, the mupen64plus-next core will crash with a segmentation fault on builds that use the libmali graphics driver. For this driver, a workaround is required that involves setting "GL_USE_DLSYM" before compiling the core.

See this issue in the mupen64plus-next core repository: https://github.com/libretro/mupen64plus-libretro-nx/issues/288

I will open a pull request that fixes this issue.

asdf288:
This fixes #1153


m4xw:
Can you PR the dlfnc include to my repo?

asdf288:
hi m4xw, I'm new to git and still having some trouble with it, it lists my linux username as co-author, adds commits that were already merged and other things. Until I get this fixed, would you mind adding it yourself? Also, the #ifndef statement can be left out I think. I've read that the preprocessor will only include it once if it detects an include guard.

m4xw:
will do that some other day, just didnt want to snatch the diff :P
Gonna write here afterwards

PR #1159, opened 2021-01-15
kernel-firmware: use copy-firmware.sh

by-gnome:
Since we are using the newest linux-firmware (kernel-firmware) we can apply this changes from LE upstream.


ToKe79:
Hi,
this PR (bc60f54bf87b8e911532f11295105df3e88a6fee) did not build correctly for all targets.
Following targets failed (2):
- Generic.x86_64
- Generic.i386

Logs and images for testing can be found here: http://nightly.builds.lakka.tv/PRs/pr_1159_000203/

PR #1172, opened 2021-02-06
create / get libretro tarball

ToKe79:
two new tools:
create_libretro_tarballs.sh: creates a tarball of already existing
folder with cloned repository for future re-use

get_libretro_tarball.sh: retrieves a tarball from a mirror in case a
specific version cannot be checked-out by the buildsystem

create_libretro_tarballs.sh should be run by the mainteners on regular
basis, so there are backups of most versions of used libretro
packages/cores

get_libretro_tarball.sh should be run by a builder in case the build
errors due to non-existent hash in cloned repository

### DIFF
--- a/create_libretro_tarballs.sh
+++ b/create_libretro_tarballs.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+# Creates .tar.xz archives of cloned libretro repositories for future re-use
+# in case original repository goes awol or is force-pushed, so a build can
+# be reproduced.
+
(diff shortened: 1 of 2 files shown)


ToKe79:
Hi,
the build of this PR (ee80a9c4192f7192c9f968a7b1d2dcb454dbfefd) finished successfully for all 20 targets.
Images for testing and logs are here: http://nightly.builds.lakka.tv/PRs/pr_1172_000213/

PR #1182, opened 2021-02-18
Adding initial support for Gameforce CHI console

davidgfnet:
This is essentially an OGA clone with a bigger (640x480) non-rotated
screen. Requires a joypad kernel patch (a copy of OGA's + small bits) and a new
DTB.

Changed a bit the way splash screens work, now you can specify proper
resolutions (instead of just vertical res) so that it works well for
cases like rotated screens.

Tweaked Retroarch to ship both drivers (DRM + GO2 DRM) and the GO2
driver to inhibit itself in case the console has a non-rotated display.
This should work for any console as long as the resolution is reported
correctly. The regular DRM driver should have 1 frame less latency!

What's not working: Console is still rotated (minor issue) and kernel
panics on poweroff :) Debugging it, but I'll need to solder the UART for
that.


ToKe79:
Hi,
this PR (80905def587ccdb158f7eb3171d268624230786e) did not build correctly for all targets.
Following targets failed (1):
- RK3328.arm

Logs and images for testing can be found here: http://nightly.builds.lakka.tv/PRs/pr_1182_000220/

PR #1201, opened 2021-03-10
lutro: export HAVE_NEON to fix build

SethBarberee:
Didn't do a fresh build on lutro and found a compile issue for `ARM` when NEON is enabled.


Ntemis:
Thanks

PR #1222, opened 2021-03-21
Fixes for RG350V: enable second MMC and Wifi

davidgfnet:
This reenables Wifi (MT7601U) that was deleted by a previous commit.
Also enables the secondary SD card reader by fixing the voltages.
Finally re-enable low power OPPs that were disabled. This brings power
savings on idle mode and low demanding emus (in particular voltages are
lower for those points, reducing power consumption significantly)


Ntemis:
Good work

PR #1235, opened 2021-04-02
Generic: enable vulkan, increase partition size

ToKe79:
minor changes also to RPi - changed VULKAN to vulkan-loader,
vulkan-headers are pulled in as dependency of vulkan-loader

### DIFF
--- a/projects/Generic/options
+++ b/projects/Generic/options
@@ -12,6 +12,8 @@
         #
         TARGET_CPU="x86-64"
         TARGET_FEATURES="64bit"
+        VULKAN="vulkan-loader"
+        SYSTEM_SIZE=1024
         ;;
       i386)
         TARGET_CPU="i686"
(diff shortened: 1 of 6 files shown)


projects/Generic/options
```diff
@@ -12,6 +12,8 @@
         #
         TARGET_CPU="x86-64"
         TARGET_FEATURES="64bit"
+        VULKAN="vulkan-loader"
+        SYSTEM_SIZE=1024
```
by-gnome, line 16: does the partition size will be increased after update?

ToKe79: no, this would require shrinking the storage, creating new storage, moving files from storage, deleting old storage expanding system, copying over system+kernel... it is easier to manually copy over files from old storage to new storage

ToKe79: what I mean - it is much easier to just burn the image to new disk, expand the storage partition on new disk and copy over data from storage partition on old disk to the storage partition on the new disk. or just to make a backup of /storage, burn new image to the same disk, restore backup (backup of course on a different medium than Lakka).

PR #1243, opened 2021-04-06
rpi4-aarch64: enable CONFIG_CIFS_WEAK_PW_HASH

dmrlawson:


Ntemis:
Why only for aarch64?

dmrlawson:
> Why only for aarch64?

It's for someone to test to see if it solves their problem. If it does we can enable it everywhere

dmrlawson:
It didn't fix the problem

PR #1247, opened 2021-04-09
Make the GPiCase actually display

SethBarberee:
Current config will give a black screen. The SPI Screen is separate from KMS/DRM so we need `fkms` instead to display.

NOTE: It's rotated 90 degrees counter-clockwise due to KMS.


ToKe79 (approved):

fdziarmagowski:
This enables the screen output, but the screen is rotated 90° left.

SethBarberee:
@fdziarmagowski Yes, that is correct. As I noted in the PR, KMS rotates it 90 counterclockwise. That is a known issue.
There isn't much I can do at the moment to fix that in Retroarch. Crux of it is https://github.com/libretro/RetroArch/issues/6770

The cores are fixed by setting the rotate option in retroarch (which I will PR later this week to set that automatically and fix screen resolution configuration)

PR #1255, opened 2021-04-11
Fix Genesis Wide .so filename

davidgfnet:
It was overwritting in some cases the wrong .so file. This enables both
emulators to be built


RobLoach (approved):

ToKe79:
Thanks - actually - I wanted to remove the file name from the target (but got distracted?), as it is totally unnecessary (I based the package.mk on non-wide genesis plus gx).

Issue #1174, opened 2021-02-07
Pi4 gpio-shutdown not working


donald-keighley:
Pi4 8GB model with Lakka-RPi4.arm-nightly-20210206-4ea79b3.img.

Using this in config.txt to power on and shutdown with a button as well as turn on the button's led.: 

`dtoverlay=gpio-shutdown,gpio_pin=3
dtoverlay=pi3-act-led,gpio=23
dtparam=act_led_trigger=default-on`

This works with raspbian and it used to work on the stable version of Lakka, so it's not a problem with how the button is wired.  The led part works with the current nightly, but the shutdown part does not.

Turro75:
I had similar issues while porting gpio joystick controls to pi4 due to changes in how gpio is initialized, while fixing the code I used to set input pullups in the config.txt as reported here: https://www.raspberrypi.org/documentation/configuration/config-txt/gpio.md

gouchi:
It is it still an issue with [latest nightly](http://nightly.builds.lakka.tv/latest/) ? Moreoever, you can use now [RPi.GPIO](https://github.com/libretro/Lakka-LibreELEC/pull/1264).

Thank you.

jmdcampos:
@gouchi , this dtoverlay line still does not work. For now, RPi.GPIO is the way to go. 
However, using config.txt is much easier and probably also requires less CPU cycles than a Python script.

Using Lakka-RPi4.arm-nightly-20210419-5a4fc0e.img

donald-keighley:
Just tried the latest nightly Lakka-RPi4.arm-nightly-20210423-d52f851.img, still did not work.

donald-keighley:
Still not working in Lakka 3.0 for Pi4.

donald-keighley:
Based on this [post](https://www.raspberrypi.org/forums/viewtopic.php?t=261452) it looks like 

```HandlePowerKey = ignore```

Should be changed to 

```HandlePowerKey = poweroff```

in the file /etc/systemd/logind.conf.

Unfortunately, I can't edit it because it's read only.

gouchi:
@donald-keighley 
Thank you for the investigation.

It seems to come from this [commit](https://github.com/libretro/Lakka-LibreELEC/commit/bf217bc814ba5489b3208019d0559717d2b5b1b9).

According to the [documentation](https://www.freedesktop.org/software/systemd/man/logind.conf.html),

[...]

PR #1264, opened 2021-04-15
Added the RPi.GPIO Python library to the RPi Project

### LINKED ISSUE #1227: [Feature request] Adding pip/RPi.GPIO module to Python
Hello,

I'm writing this feature request to possibly add the pip module for Python. I think Lakka would greatly benefit from extended Python support. The best way to do that without adding size overhead to Lakka and minimizing average user error would be to include the pip module that allows you to install nearly any Python module, giving you access to what you need if you need it, but using barely any storage if adding no other modules.

This would most importantly help with GPIO and overlay features, but other use cases can easily be thought of. Upgrading to Python 3 could also be a good idea, but is off-topic, and as a non-expert about Python I'm not sure how much complexity it would add to Lakka.

Thanks in advance for considering it !

DevilBlackDeath:
Following my feature request #1227 , after some tests is proves nearly impossible to have pip in Lakka. The efforts required to make it work makes adding python packages on a per-user basis easier. However RPi.GPIO is still a vital  package as it allows a lot of stuff, including safe shutdown scripts for cases with power buttons. While there are a lot of precompiled RPi.GPIO floating around, these only account for arm compilations, not aarch64. For such a core Pi Python component, I feel like having it compile with Lakka would be a benefit.

I've forked and am working on a safe shutdown script that could serve as the basis for multiple cases for Lakka users (mine is for Retroflag's but it would be easy to fork and adapt it).

Fix #1227

RPi.GPIO package upgraded to SPDX format. Being unsure whether credits should go to Stephan Raue, the LibreElec team, both and/or the Lakka team, I templated one of the SPDX files,

### DIFF
--- a/packages/python/utils/python-raspberry-gpio/package.mk
+++ b/packages/python/utils/python-raspberry-gpio/package.mk
@@ -1,20 +1,6 @@
-################################################################################
-#      This file is part of OpenELEC - http://www.openelec.tv
-#      Copyright (C) 2009-2016 Stephan Raue (stephan@openelec.tv)
-#
-#  OpenELEC is free software: you can redistribute it and/or modify
-#  it under the terms of the GNU General Public License as published by
-#  the Free Software Foundation, either version 2 of the License, or
-#  (at your option) any later version.
-#
-#  OpenELEC is distributed in the hope that it will be useful,
-#  but WITHOUT ANY WARRANTY; without even the implied warranty of
-#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#  GNU General Public License for more details.
-#
-#  You should have received a copy of the GNU General Public License
-#  along with OpenELEC.  If not, see <http://www.gnu.org/licenses/>.
-################################################################################
+# SPDX-License-Identifier: GPL-2.0-or-later
+# Copyright (C) 2009-2016 Stephan Raue (stephan@openelec.tv)
+# Copyright (C) 2019-present Team LibreELEC (https://libreelec.tv)
 
 PKG_NAME="python-raspberry-gpio"
 PKG_VERSION="0.7.0"
--- a/projects/RPi/options
+++ b/projects/RPi/options
@@ -143,5 +143,10 @@
       DRIVER_ADDONS=""
     fi
 
+  # additional packages to install:
+  # Space separated list is supported,
+  # e.g. ADDITIONAL_PACKAGES="PACKAGE1 PACKAGE2"
+    ADDITIONAL_PACKAGES="python-raspberry-gpio"
+
   # debug tty path
     DEBUG_TTY="/dev/console"


DevilBlackDeath:
I've now finished my own safe shutdown script for the retroflag case. merging this request would let any retroflag NESPI 4 (and afaik, any other retroflag case) Lakka users to use my safe shutdown/power cutoff script. This would also allow any user of GPIO-enhanced cases to use my scripts as a base for similar functionalities when running Lakka.

jmdcampos:
@DevilBlackDeath , I tested your script. It caused a kernel panic when pressing the shutdown button. This is easily fixed by removing the dtoverlay line your installation inserted in config.txt, otherwise we get two processes reacting to the shutdown button and causing a conflict. 
I will try to confirm if #1174 is still an issue or not. I'll have to remove the script before that, to avoid this conflict.
PS: Using RPi4.arm build with a custom wired case.

DevilBlackDeath:
@jmdcampos Does the Nespi case still is able to cutoff power ? (listen to fan or look for the LED if the case is still open). As far as I know dtoverlay is required for actual power cutoff. It causing a kernel panic could be a Lakka issue. I'll look into that, but as far as I know power cutoff when using Safe Shutdown won't work without the dtoverlay (it's in charge of actually giving the GPIO signal to the case board once shutdown is complete). If the panic doesn't bother you in and of itself you could simply use the power button as the whole point of the safe shutdown script is to be able to press the power button without manually shutting down.

DevilBlackDeath:
@jmdcampos 
After doing some reading here : https://www.raspberrypi.org/forums/viewtopic.php?t=143519
and there : https://github.com/raspberrypi/linux/issues/3372

It would seem the kernel panic is actually an intended caution warning. Since the kernel is expecting the power to be cut off and it actually isn't it throws the warning. As such, manually shutting down on a Retroflag case with safe shutdown enabled is considered "abnormal" behaviour. As indicated in my previous message you should then only use the power button to turn the pi off (BUT make sure safe shutdown is turned ON beforehand, otherwise you'll just cut power off to the pi while the OS is running, which is highly unadviseable). Also please note that if you use a non-retroflag case with similar behaviour you may have to adjust a couple of things (namely the GPIO pins in the dt-overlay and Python script).

Hope I could be helpful :) Tell me if there's anything else !

jmdcampos:
Hi,
My case has no fancy circuitry to cut power. There is  just a LED and a momentary button, used to trigger the shutdown procedure in software. Then the Pi simply powers down...

DevilBlackDeath:
@jmdcampos The Pi can't power down on its own. If it powers down it means the case is doing it somehow (best way to check that is to see if the fan is still running). Since you do have a button on the case, if the fan actually shuts down, it's very likely the case has the circuitry to power down the Pi.

Does the Pi connect through GPIO ? You likely can't use my script as is but it's fairly easy to adapt. You should change the POWER and LED pins in the Python script (and you can actually fully remove the Reset pin and reset functions if you don't have a reset button). Then you should change the dt-overlay power-off pin to the pin actually in charge of powering down the Pi. That's the "harder" part. A lot of cases don't document these very well, and if they're as "amateur" as Retroflag when it comes to this particular thing, it means you'll have to hunt down a working shutdown script made by someone else and find which pin does it. There's also very likely a way to find out by looking at the case and GPIO wiring which pin needs to be used, but since there was a galore of scripts for Retroflag, finding the pin was fairly easy so I didn't take the time to do that.

Tell me what case you're using and I'll try to see what I can find ;)

jmdcampos:
Well, like I said before, my case has no power circuitry at all. It's just a wooden box with some holes, a power LED and a shutdown button connected to GPIO pins, and that's all. These work as intended with your script: the Pi boots when the button is pushed (GPIO 3), and the LED lights up, too (GPIO 14). And if the button is pushed again, the OS shuts down and the LED is put off. Since the power cord is still connected, then yes, the Pi is not really powered off. But it is not working, either.... No fan yet, and that would probably be connected via transistor circuit to another GPIO pin.

DevilBlackDeath:
@jmdcampos Oh okay, well then yes, you need to get rid of the gpio-poweroff indeed and not much more you can do. You can always fork my work if you want and make it more generic and compatible with case that have no power cutting ;)

jmdcampos:
@Ntemis 
No need to remove RPi.GPIO. The said kernel panic was caused not directly by RPI.GPIO, but by an external script which uses it.  But @DevilBlackDeath  and me both agree this was normal. That external script only needs some custom editing to fit each user's needs, otherwise it works. GPIO.RPi is a great new feature for Lakka, as it enables custom external cases to finally work properly.

bfocht:
Any word on when GPIO is will be re-enabled?

DevilBlackDeath:
@Ntemis This is perfectly safe and is a package included in LibreELEC, I only had to reenable it to make it active. As explained above, the kernel panic is expected behaviour for my script. No risk to data afaik (and from testing I've seen no corruption). You can safely reenable it ;)

Ntemis:
Done, thanks for confirming.

DevilBlackDeath:
You're welcome !

PR #1273, opened 2021-04-22
Rpi kernel update to hopefully fix composite out

dmrlawson:


eugene-s-nesdev:
Ntemis comment:
> Report this to raspberry pi kernel repo so we can fix this in the future

I confirm this problem on RPi3B+ with lastest "Lakka-RPi3.aarch64-nightly-20210724-58dafbe.img" 
- i've installed fresh lastest build and tested it on 1920x1080 display via HDMI - all works good.
- added this lines to config, then plug RPI to CRT-TV via 3.5mm out:
```
hdmi_force_hotplug=0
hdmi_ignore_hotplug=1
sdtv_mode=16
```
Signal is present (i see deep dark-gray backgroung with some noise) but there are no picture.

Lakka 2.2.x/2.3.x works excellent on CRT with same settings

PR #1279, opened 2021-04-28
yabasanshiro: build for aarch64

dmrlawson:
I'm not actually sure what the difference is between this and yabause but apparently some people care about it and it seems to build for rpi4.


Ntemis:
No comparison ☺️ just test it

PR #1298, opened 2021-05-16
Rpi kernel update

dmrlawson:
I've merged `bcm2835-driver` into `bcm2835-bootloader` as they're just different parts of the raspberry/firmware repo, doing it this way means we can update them together.

I think it should be renamed `rpi-firmware` but let's check this works first.


Ntemis:
you havent test this switch yet?

Ntemis:
you haven't test this switch yet?

PR #1303, opened 2021-05-19
rpi: make bcm2835-driver use the same version as bcm2835-bootloader

dmrlawson:
This is what my previous PR was trying to solve. It gets files from the same package, but we fetch it from libreelec currently meaning we're limited in which commits we can pin to.

This change just fetches the exact same package and the exact same version as bcm2835-bootloader (https://github.com/libretro/Lakka-LibreELEC/blob/master/packages/tools/bcm2835-bootloader/package.mk#L6). In the future we can merge them.


Ntemis:
Did you test this? Because we are not in state we can do experiments atm

PR #1308, opened 2021-05-20
Remove the RG351P/V rumble patch

davidgfnet:
It produces random kernel panics. Until I figure out what the bug is,
disable this.


Ntemis:
Can we release lakka for these handhelds  if this bug is fixed?

PR #1325, opened 2021-05-24
L4T: Fix Vulkan Renderer

GavinDarkglider:
This commit fixes the vulkan renderer on L4T platform builds(Nintendo Switch)


tabnk:
Tested on Switch but only shown black screen with a CROSS cursor.

PR #1343, opened 2021-05-26
noobs: replace slides, modify partition script

ToKe79:
forgot to update these files when porting to LE9.x
buildbot: [dnb]


by-gnome (approved):
Better late than never.

PR #1356, opened 2021-05-31
swanstation: fix package.mk

dmrlawson:
It was renamed in the repo so it broke when we updated


Ntemis:
your fix is half done bud

dmrlawson:
How so? You want to rename the package as well?

Ntemis:
It makes perfect sense, don't you think?

dmrlawson:
Overtaken by https://github.com/libretro/Lakka-LibreELEC/pull/1357

PR #1364, opened 2021-06-02
ppsspp: don't use system ffmpeg as recommmended

ToKe79:


m4xw:
Very important actually!
Theres a few changes that they maintain and newer ffmpeg has breaking behaviour in versions from a year ago to my knowledge.
It never worked right without ppsspp-ffmpeg on my Switch port.

PR #1371, opened 2021-06-05
retroarch: build with gles 3 minor version support [WIP]

dankcushions:
See https://github.com/libretro/RetroArch/pull/11181

This is UNTESTED, so feel free to close and consider it as advisory :) I don't have a lakka dev environment/spare SD. Just something I was meaning to eventually do after above PR was merged. It's only really relevant for things like swanstation that actually use the API to determine GLES version support. see https://github.com/stenzek/duckstation/issues/694. You can see if it's working via the presence of GLES 3.x items in a verbose log (regardless of what core retroarch is running).

I don't know if any of these other devices need updating. I imagine a few are gles3.x


ToKe79 (approved):
LGTM

PR #1378, opened 2021-06-07
[ISSUE-1355] rpi: update wifi firmware

dmrlawson:


kellenqian:
seems this update fix the pi400 wifi/bt issue

PR #1382, opened 2021-06-09
swantstation: default to sw renderer when no OGL

ToKe79:
This patch makes SwanStation to compile with default  GPU renderer set to Software, so it can be started on platforms without OpenGL support. For platforms with OpenGL support this patch is reversed during post-patch.

### DIFF
--- a/packages/libretro/swanstation/package.mk
+++ b/packages/libretro/swanstation/package.mk
@@ -50,6 +50,14 @@ PKG_CMAKE_OPTS_TARGET="-DCMAKE_BUILD_TYPE=Release \
                        -DUSE_DRMKMS=ON \
                        -DBUILD_LIBRETRO_CORE=ON"
 
+# Compile with software renderer as default for devices not suporting OpenGL
+if [ "$PROJECT" = "Generic" -o "$DEVICE" = "RPi4" -o "$DEVICE" = "Switch" ]; then
+  PKG_PATCH_DIRS=""
+else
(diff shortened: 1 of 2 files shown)


packages/libretro/swanstation/package.mk
```diff
@@ -50,6 +50,13 @@ PKG_CMAKE_OPTS_TARGET="-DCMAKE_BUILD_TYPE=Release \
                        -DUSE_DRMKMS=ON \
                        -DBUILD_LIBRETRO_CORE=ON"
 
+post_patch() {
+  # revert default to OpenGL renderer where supported - Generic + RPi4
+  if [ "$PROJECT" = "Generic" -o "$DEVICE" = "RPi4" ]; then
```
plaidman, line 55: switch also uses opengl. `-o "$DEVICE" = "Switch"`

dmrlawson:
Another option is to use `PKG_PATCH_DIRS` like [the kernel package](https://github.com/libretro/Lakka-LibreELEC/blob/master/packages/linux/package.mk#L36).

E.g. put it in a `no_opengl` directory and set `PKG_PATCH_DIRS="no_opengl"` for the desired platforms.

It saves a bit of work in not patching and un-patching, but the current solution is nice

plaidman:
I thought patching and unpatching was kinda hacky tbh - I like dmrlawson's idea.

plaidman (approved):

PR #1385, opened 2021-06-10
L4T: enable melonds

plaidman:
this enables MelonDS for aarch64 devices and enables JIT for switch

### DIFF
--- a/distributions/Lakka/options
+++ b/distributions/Lakka/options
@@ -511,7 +511,7 @@ DISTRO_SRC="http://sources.libreelec.tv/$LIBREELEC_VERSION"
       LIBRETRO_CORES+=" mame2000  fbalpha2012 "
     fi
   fi
-# add melonds to x86_64
-  if [ "$ARCH" = "x86_64" ]; then
+# add melonds to x86_64 and L4T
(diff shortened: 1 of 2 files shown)


distributions/Lakka/options
```diff
@@ -511,7 +511,7 @@ DISTRO_SRC="http://sources.libreelec.tv/$LIBREELEC_VERSION"
       LIBRETRO_CORES+=" mame2000  fbalpha2012 "
     fi
   fi
-# add melonds to x86_64
-  if [ "$ARCH" = "x86_64" ]; then
+# add melonds to x86_64 and L4T
+  if [ "$ARCH" = "x86_64" -o "$ARCH" = "aarch64" ]; then
```
ToKe79, line 515: probably change `"$ARCH" = "aarch64"` to `"$DEVICE" = "Switch"` - RPi4 is also aarch64 and it fails to build.

plaidman: my guess is there's something wrong in the makefile for `platform=rpi4_64`, so I just went back to using `platform=unix` for that one. seeing as it compiled for rpi3, it's probably good.

ToKe79 (approved):

PR #1392, opened 2021-06-13
flycast: enable OIT for OpenGL platforms

plaidman:


ToKe79 (dismissed):
iirc adds an option to the core settings (default off), so should not introduce any regression.

plaidman:
no updates just rebased onto master

ToKe79 (approved):

PR #1395, opened 2021-06-16
linux: Enable CONFIG_BLK_DEV_DM on Generic

### LINKED ISSUE #1381: Feature request - Contain dm-mod.ko by default.
Hi,
With [Ventoy](https://github.com/ventoy/Ventoy) we can just copy the lakka.img to the USB drive and boot it.
But Ventoy need the `dm-mod.ko` which is not included in the img file.
Can we build and contain the `dm-mod.ko` in the img file by default in the future release?

Thanks!

gouchi:
- Fix #1381


ToKe79 (approved):
failed platforms are not related to this change

PR #1399, opened 2021-06-22
rpi3: fix analog audio output

dmrlawson:
Possible fix for 3.5mm audio output.

I've not tested this but would like to before merging.


dmrlawson:
theknob on discord claims this works and doesn't affect the ability to use hdmi audio so I suggest merging

PR #1405, opened 2021-06-26
Update vulkan-tools, vulkan-loader, vulkan-headers and build vulkan-tools for rpi4

dmrlawson:


dmrlawson:
I don't think my change should affect AMLG12 in any way

dmrlawson:
^ rebased

PR #1414, opened 2021-07-02
Fix Dual Sense FF and LEDS by updating patches to v6.0

coldspark29:
Version of Dual Sense patches from 2021-02-07. 

See https://patchwork.kernel.org/project/linux-input/list/?series=&submitter=162701&state=*&q=&archive=&delegate=

Fixes: 

- Force feedback
- LEDs now show player number 

Not the latest and greatest patches, but everything works. Good enough to leave it like this until the next LTS kernel is released.


coldspark29:
One of the patches was missing and everything is fixed now. The build finished successfully.

PR #1421, opened 2021-07-08
Fix some host tools with newer gcc/glibc

davidgfnet:
This contains version bumps (borrowed some stuff from upstream) for fakeroot and elfutils.
Building with gcc11 and glibc2.33 seems to cause issues due to some breakage in newer glibc header.


davidgfnet:
Please wait for the build to finish, I tested this locally but might break other platforms. Thanks!

PR #1435, opened 2021-07-25
retroarch: revert resampler changes (patch)

by-gnome:
Still sound noisy after savestave loading on iMX6.
Patch modified after libretro/RetroArch@3f00d2f041abcd444f0472e7b05ee87ff107c948.


Ntemis:
@twinaphex just a fyi for future usage

ToKe79:
@by-gnome other platforms were not affected? I guess we could use the patch then only for imx6. please confirm if only imx6 is affected.

by-gnome:
I tested Generic, RPi and iMX6 - of these three only iMX6 is affected.
Ok, I will move it to iMX6 project.

PR #1467, opened 2021-08-21
gcc : add libstdc++.a to sysroot

gouchi:
- Will be needed for the Play! core


gouchi:
Will submit one PR with #1468

PR #1468, opened 2021-08-21
play : update play core

gouchi:
- Remove patch which is not needed anymore need #1467
- Disable glew 
- Update to latest commit
- Enable the core for all arch


gouchi:
Will submit one PR with #1467

PR #1475, opened 2021-08-27
Add dosbox-pure core

asdf288:
I have tested this on armv7, armv8 and x86-64 builds.


Ntemis:
hmm i have already added this a few hours back

asdf288:
Oh, never mind then. Yours was better anyway, I forgot to add the core to the options file.

PR #1486, opened 2021-09-06
Add support for RPi4 PiBoy DMG

dmrlawson:
For https://github.com/libretro/Lakka-LibreELEC/issues/1060

This adds all the bits that @duduke worked on, battery state still todo.


Ntemis:
Please add it to the build devices script too
https://github.com/libretro/Lakka-LibreELEC/blob/master/build_all.sh#L90

PR #1488, opened 2021-09-07
lakka-update: Use HTTPS and fetch last update correctly

ItachiSan:
While trying to update my NOOBS RPi3 via `lakka-update`, I noticed some
issues in the script.
I did a few experiments in the SSH session and reached a few noticeable
upgrades, available in this commit.

Namely:
- By switching from `head` to `tail` in parsing the file `.index`,
  the script now fetches correctly the latest version.
- The mirror has been updated to use HTTPS, which is overall safer
  and better.

You can verify that the latest files are fetched via something alike
this:

    for distro in `wget -O - https://le.builds.lakka.tv/ -q | grep '<tr>' | sed -E 's|.*href="([A-Za-z0-9._-]+)/".*|\1|' | tail -n +2`
    do
      echo "# Device: $distro"
      wget -q -O - https://le.builds.lakka.tv/$distro/.index | tail -1
      echo
    done


ItachiSan:
While I was not able to update `lakka-update` in my SSH session, I essentially ran all the commands in the shell script manually step by step.
If you would like me to, I can rebuild the package and test it locally; if so, can you point out how to test out a single package? 😄 
Thanks in advance!

ToKe79:
Thank you for your findings.

This looks like fault on my side - I generated the `.index` files using `ls *.tar > .index` instead of `ls -t *.tar > .index` (I did not see a difference in RetroArch - looks like RetroArch sorts the file list from `.index` alphabetically). Hereby the `.index˙` files are fixed (most recent file at top) - no need for the `head`/`tail` swap.

As per the change of protocol to https - this is kind of an enhancement that the user can be assured, that they get the file from the proper source. But http works as well, was there any additional reason for the change?

However, as you poked this package - can you also modify the package.mk so that it replaces `le.builds.lakka.tv` with `nightly.builds.lakka.tv/.updater` in the `lakka-update`-script when `NIGHTLY_BUILD` is set to `yes`? You can add it in the `makeinstall_target()` function.

ItachiSan:
Hey Tomas,
> 
> This looks like fault on my side - I generated the `.index` files using `ls *.tar > .index` instead of `ls -t *.tar > .index` (I did not see a difference in RetroArch - looks like RetroArch sorts the file list from `.index` alphabetically). Hereby the `.index` files are fixed (most recent file at top) - no need for the `head`/`tail` swap.
> 

That's good; is this change now permanent (AKA next generations of `.index` will be like this)?

>
> As per the change of protocol to https - this is kind of an enhancement that the user can be assured, that they get the file from the proper source. But http works as well, was there any additional reason for the change?
> 

HTTPS ensure that the correct host is addressed and avoids MITM attacks.
In a really far-stretched but possible scenario, one could setup a malicious server behaving as the Lakka server and allowing for malicious system updates.
Having HTTPS in the first place would verify that the HTTPS communication is valid via the certificate.

> However, as you poked this package - can you also modify the package.mk so that it replaces `le.builds.lakka.tv` with `nightly.builds.lakka.tv/.updater` in the `lakka-update`-script when `NIGHTLY_BUILD` is set to `yes`? You can add it in the `makeinstall_target()` function.

I can do that, however I have a few questions:
* Is `NIGHTLY_BUILD` an OS variable or is defined at build time?
* Is there any example of packages using variables at build-time? If I have a pseudo-shell, I can modify stuff easily (I do that when packaging for Archlinux)

ToKe79:
Hi Giovanni,

> That's good; is this change now permanent (AKA next generations of `.index` will be like this)?

Yes, I have corrected the script that generates the `.index` files.

> I can do that, however I have a few questions:
> 
> * Is `NIGHTLY_BUILD` an OS variable or is defined at build time?

Yes, it is build time env usually passed to the `make` command, however, the correct name is `LAKKA_NIGHTLY` (sorry for the confusion), e.g. `LAKKA_NIGHTLY=yes PROJECT=RPi DEVICE=RPi4 ARCH=aarch64 make image`.

> * Is there any example of packages using variables at build-time? If I have a pseudo-shell, I can modify stuff easily (I do that when packaging for Archlinux)

The only example (I know of) is in `retroarch` package:

https://github.com/libretro/Lakka-LibreELEC/blob/b753512546518dc8f94439a182b16cf8ecbfaeb0/packages/libretro/retroarch/package.mk#L156-L159

Here it sets a value for `HAVE_LAKKA_NIGHTLY` to be passed to the make command for RetroArch, so RetroArch uses correct DEFINES to use different URL in the RetroArch build of the nightly images, so users can update between nightly builds instead to a release - which would be in many cases a downgrade.

In your change make sure to modify the `$INSTALL/usr/bin/lakka-update` file.

packages/lakka/lakka-update/package.mk
```diff
@@ -36,6 +36,12 @@ make_target() {
   :
```
ToKe79, line 36: Can you also remove `PKG_DEPENDS_TARGET="..."` (the whole line 25) - it does not need the `toolchain` to be built before the file is copied. Sorry, I did not see this before.

ToKe79 (changes requested):
Thank you, just one more request - see in comments.

ItachiSan:
Hi @ToKe79 ,

Got some time again!

I read through the guide here:
https://github.com/LibreELEC/LibreELEC.tv/blob/master/packages/readme.md
which I would recommend linking somewhere as it might help others :smile:
It is also available in this repo  at: packages/readme.md :wink: 
After that, I was able to build stuff:
```bash
$ LAKKA_NIGHTLY=yes ./scripts/build lakka/lakka-update
UNPACK      lakka-update
BUILD      lakka-update (target)
    TOOLCHAIN      manual
$ grep ^MIRROR build.Lakka-Generic.x86_64-3.4-devel/lakka-update-0/lakka-update.sh 
MIRROR=https://nightly.builds.lakka.tv/.updater
$ rm -rf build.Lakka-Generic.x86_64-3.4-devel/
$ ./scripts/build lakka/lakka-update
UNPACK      lakka-update
BUILD      lakka-update (target)
    TOOLCHAIN      manual
 $ grep ^MIRROR build.Lakka-Generic.x86_64-3.4-devel/lakka-update-0/lakka-update.sh 
MIRROR=https://le.builds.lakka.tv
```
I do believe the output is as it should :wink:

I've reverted `tail` to `head` and added the correct `sed` replacement in `pre_makeinstall_target()`, which does not change sources in `packages` but in `build.OS-COMBO`.

packages/lakka/lakka-update/package.mk
ItachiSan: Hi again! It seems I forgot to press the "Comment" button :laughing: 
I can fix that a bit later, but sure!

ToKe79 (approved):

Issue #1358, opened 2021-05-31
Computer Module 4 & Lakka 3.0 issue with the USB 2.0


kelroy1990:
Hi guys, i am testing the lakka 3.0 release on the Computer Module 4 and the raspberry pi IO board.

The display and the HDMI its working well as i see the image correctly. But we have a few issue with the USB.

Computer Module 4 has his USB off by default, so we need to use the **dtoverlay=dwc2,dr_mode=host** to activate it. But on Lakka 3.0 it seems to dont have any kind of effect, so i cannot control the SO.

I see that Lakka have the dwc2 overlays on the /boot/overlays.... so i cannot figure where the problem is.

Its somebody testing it like me?.

Next Steps:
-Configure SSH with a pi4 destkop to have internet and test it on the CM4

Regards, Carlos.

shawhall:
Did this get fixed?  I see the issue was closed but no comments were left.

Ntemis:
You didnt update the issue after all the stable releases and no one on the team has the Cm4 so no idea if it was fixed or not but we cant walk blindly. Try v3.2 and let me know though.

shawhall:
Just tried v3.2 today and it does not work.

kelroy1990:
Hi guys, tried to and it does not work. If the lakka team need one, i could send it to they, just contact with me.

I was cheking the kernel build and the dwc2 module seems to be active to compile, i need further work to see where its the fail.

dmrlawson:
Hi, I think this should work now in the latest nightlies if you enable it in distroconfig.txt

Issue #1188, opened 2021-02-28
Lakka not enabling USB on CM4-IO Board


jmacato:
> Which version of Lakka are you using?

Nightly (Lakka-RPi4.aarch64-nightly-20210227-e925215)

> What system hardware are you using?

Raspberry Pi Compute Module 4 (with eMMC) + CM4 IO Board.

> What did you do?

So i flashed the nightly build into the CM4 and also added the dtoverlay option needed to activate its usb hub. It worked on RPi OS latest but whatever i tried, it still didn't enable the USB hub in Lakka for some reason.

> What did you expect to happen?

Being able to use USB.

> What happened instead?

Didnt work.
 
 
 I have a hunch that this could be fixed by including the latest raspberry pi firmware files? Not sure though.

SethBarberee:
Is this still occurring for you? Firmware has been updated since the opening of this.

othermod:
I just tried with the latest nightly build. USB is still not working, even with the dtoverlay=dwc2,dr_mode=host added to config.txt

jmacato:
@SethBarberee i can confirm @othermod 's findings.

yikesmeguts:
Although I do not have the CM4 IO board, I do have a CM4 Lite and a GPiMate+  CM4.  Using the Pi4.arm Lakka build, and applying the GPiCase-specific settings, Lakka will boot on the GPiMate+ CM4, but the system will not respond to any input from the dpad.  Having used the same GPiCase-specific settings to configure RetroPie, I can confirm the hardware does work properly when dtoverlay=dwc2,dr_mode=host is defined in config.txt for RetroPie, but not for Lakka on the CM4. I do realize that Lakka and RetroPie are vastly different systems, but the above fix doesn't seem to apply in this case.

On an unrelated note, I also have a GPiMate+ CM3 and Lakka runs beautifully on the CM3.  Once the screen orientation issue has been fixed, it will be the perfect OS for the GPiCase, though I would prefer to use the CM4.

jmacato:
@Ntemis may i ask why is this closed? The issue isnt fixed yet i think

Ntemis:
Sure, we don't have the device so I do not think a blind fix will be durable in the long run

dmrlawson:
Hi, I think this should work now in the latest nightlies if you enable it in distroconfig.txt

PR #1501, opened 2021-09-15
Add support for RPi4-RetroDreamer

dmrlawson:
This was done with help from @davidgfnet.

Everything seems to work as expected.

Still to work on:
 - battery percentage (we're exploring whether there's hardware for that)
 - screen brightness (as above)
 
This should fix https://github.com/libretro/Lakka-LibreELEC/issues/1358 and https://github.com/libretro/Lakka-LibreELEC/issues/1188 also since we add the required kernel config.


ToKe79:
GPICase is OK, tested manually.

ToKe79 (approved):

PR #1505, opened 2021-09-24
La frite

devegoo:
Hi , i want to add support for Libretech LaFrite 512mb and 1gb and some sbc with this pr but i don't have special possibilities to check is that pr do not broke current working code (devices support) in Lakka
there is testing img for LaFrite but some user want to add audio over hdmi /broken currently...

https://drive.google.com/drive/folders/13kH_QZ8Y0kRlnZ8DQnGlPEXHjYBs-DO3?usp=sharing
https://forum.loverpi.com/discussion/1216/lakka-for-both-version-lafrite-512mb-1gb


devegoo:
hi , after updating my testing image to nightly from 26.09.2021 booting stopped on "starting kernel", i don't know where is issue, this happen on lafrite 512mb , in libreelec in kernel config cma=128 must be set, also in upgrating process i seen some new procedures at the end  this do something with boot.scr and there is probably problem, lafrite have uboot programmed on spiflash not like most sbc in image...

ToKe79:
HI, thank you for your work, but this device will be added in the next major release of [Lakka - 4.x](https://github.com/libretro/Lakka-LibreELEC/tree/Lakka-v4.x), which will be based on the LibreELEC 10.

PR #1520, opened 2021-09-30
Add RG351MP

davidgfnet:
This adds support for RG351MP (which is a very confusing name since variants M and P already exist!)
It is based on the RG351V configuration (with a 640x480 screen) and a hybrid joypad borrowed from Odroid Super


zongyuanliu-bio:
Did you try new firmware again? I bought RG351MP, why it can not even boot? It crash after I inserted the flashed SD card.

PR #1533, opened 2021-10-17
[WIP] Minor/Final 3.x Switch Updates

GavinDarkglider:


Ntemis:
It would be preferable to push the switch controller configs here https://github.com/libretro/retroarch-joypad-autoconfig with a pr.
Is cleaner and other switch distros can be benefited by this as well.

PR #1569, opened 2021-12-23
retroarch: drop patches that are now upstream

asdf288:
It seems that these patches are now upstream and fail to apply.


RobLoach (approved):
Thanks for checking these out :+1:

PR #1575, opened 2022-01-08
libretro: play: Rollback for older CMake

RobLoach:
The updated version of Play! requires CMake version 3.15, which is not yet available on our infrastructure. This change rolls back to the previous version of Play!.

### DIFF
--- a/packages/libretro/play/package.mk
+++ b/packages/libretro/play/package.mk
@@ -19,7 +19,7 @@
 ################################################################################
 
 PKG_NAME="play"
-PKG_VERSION="695cb21"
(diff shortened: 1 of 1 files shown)


gouchi:
The commit to change CMake version was [1321e2b](https://github.com/jpd002/Play-/commit/1321e2bf307d8acd96eab16ddc8f60f3c888828a). So we can update at least to [30a3978](https://github.com/jpd002/Play-/commit/30a3978701a16d01c54f8b15f615ec99df64551f).

RobLoach:
Nice! Good find.

PR #1576, opened 2022-01-08
libretro: tic80: Use the Release CMake build type

RobLoach:
Using `CMAKE_BUILD_TYPE=Release` will squeeze a bit more performance out of the build.


joshgoebel:
Understatement of 2022: 

> a bit more performance

PR #1592, opened 2022-01-29
scummvm: bump to fix freeze on ARM

spleen1981:


gouchi:
Hi,

Thank you for the PR. We might consider this [PR](https://github.com/libretro/Lakka-LibreELEC/pull/1593) to fix it.

spleen1981:
Yes it is covered in #1593. In this case this can be closed.

PR #1594, opened 2022-01-31
jumpnbump : add  jumpnbump to Libretro cores

loki666:
```
jumpnbump: add jumpnbump core
```

### DIFF
--- a/distributions/Lakka/options
+++ b/distributions/Lakka/options
@@ -228,6 +228,7 @@ DISTRO_SRC="http://sources.libreelec.tv/$LIBREELEC_VERSION"
 " hatari "\
 " higan-sfc "\
 " higan-sfc-balanced "\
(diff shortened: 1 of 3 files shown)


distributions/Lakka/options
```diff
@@ -284,6 +284,7 @@ DISTRO_SRC="http://sources.libreelec.tv/$LIBREELEC_VERSION"
 " xrick "\
 " yabasanshiro "\
 " yabause "\
+" jumpnbump "\
```
gouchi, line 287: May you try to keep the alphabetic order ;)

Thank you.

loki666: Arf didn't noticed... I'll change that

loki666:
should I also make a pr for Lakka-v4.x branch ?

PR #1600, opened 2022-02-07
Latest Lakka Switch Patches

GavinDarkglider:
This patch set includes:

1. Add CEC support, can be disabled via uenv.txt
2. Fix pulseaudio/alsa settings to fix issues with alsathread on dock, also switches default audio on switch back to alsathread.
3. Downgrade GCC from 10 to 9 in L4T builds to work around undefined symbol: __aarch64_ldadd4_acq_rel issue introduced in changes to linker scripts in gcc 10+.
4. Fix updater in switch builds.


Ntemis:
ok to merge?

GavinDarkglider:
Should be good to merge, but vudiq likes to review most things, so I didnt want to push directly.

ToKe79 (approved):

PR #1603, opened 2022-02-09
update jumbnbump core git ref

loki666:


RobLoach (approved):
The commit messages usually take the format of `jumpnbump: Do something really cool` , but I think this is fine.

Issue #1554, opened 2021-11-24
WiFi driver listed null, still connected to Internet


mbc3:
Under the drivers section, my WiFi driver is listed as null, and I am unable to select anything else.

I SSH'd into my Pi4 and set up my network there - and ip addr show + pinging confirms that I have access to the Internet and am on my network.

I'm not sure if this is merely a graphical display bug or if it's indicative of anything else - I'm unable to download thumbnails, for example.

RobLoach:
Confirmed this is fixed. Tested on a RPi3.

PR #1611, opened 2022-02-19
restore WIFI drivers

loki666:
a change in retroarch between retroarch v1.9.13.2 and v1.9.14 changes how wifi drivers are compiled in retroarch
adding HAVE_WIFI fix the issue

### DIFF
--- a/packages/libretro/retroarch/package.mk
+++ b/packages/libretro/retroarch/package.mk
@@ -161,9 +161,9 @@ pre_make_target() {
 
 make_target() {
   if [ "$DEVICE" = "Switch" ]; then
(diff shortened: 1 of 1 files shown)


loki666:
probably fix #1554

gouchi:
In the [Makefile.common](https://github.com/libretro/RetroArch/blob/master/Makefile.common#L802-L812) we have

```
ifeq ($(HAVE_LAKKA), 1)
   OBJ += network/drivers_wifi/connmanctl.o
   OBJ += misc/cpufreq/cpufreq.o
endif

ifeq ($(HAVE_WIFI), 1)
   OBJ += network/wifi_driver.o
   OBJ += tasks/task_wifi.o
   OBJ += network/drivers_wifi/nmcli.o
   DEFINES += -DHAVE_WIFI
endif
```
So we should try indeed with `HAVE_WIFI` also.

[Source commit](https://github.com/libretro/RetroArch/commit/71b30d78468fa7c7955d967b1579b7224c6e3a72)

loki666:
tested on RPi3

PR #1617, opened 2022-02-24
Fix recording/streaming in Lakka 4.x builds Few other minor fixes.....

GavinDarkglider:
Title says it all......

Tested with standard Twitch config from: https://docs.libretro.com/guides/recording-and-streaming/#lossy-and-flexible-coding-config

```
vcodec = libx264
acodec = libmp3lame
pix_fmt = yuv420p
scale_factor = 2
threads = 3
video_crf = 25
video_preset = superfast
video_tune = animation
audio_global_quality = 75
sample_rate = 44100
format = flv
```

this is that config ^

Place in /storage/.config/retroarch/records_config/Twitch.cfg

Then load it as the default record config option in recording options in the menu.


bonswouar:
@GavinDarkglider I don't know if I did something wrong, but I did build it from 4.x sources (Git Version `b71be7e`), used the sample config file and I'm still having the segfault

PR #1623, opened 2022-03-03
Cleanup decoder/encoder stuff for L4T Builds, update some things, fix GLES building, but keep GL build. Other Minor cleanups.

GavinDarkglider:
1. Update encoder/decoder to newest version of nvv4l, which has a 70% speed increase at 1080p
2. Add Support for vulkan to ffmpeg package if board supports it. Does it help anything? no idea, but why not, since we already have vulkan support, might as well try to leverage it where ever we can.
3. Add vitaquake3(Quake 3(Mostly broken in my testing)) and Boom3(Doom 3(Also mostly broken in my testing)) cores.


GavinDarkglider:
Changed to WIP because CTC updated encoder/decoder stuff and finalized. I need to rebase the new set of patches.

PR #1661, opened 2022-03-24
libretro update

### LINKED ISSUE #1667: Genesis Plus GX - Only the "Wide" variant available
Hello!

I'd like to report that, as of version 4.0, the regular Genesis Plus GX core is missing from Lakka, while only the Wide variant core remains. Browsing through the available cores, I can see that the regular Genesis Plus GX is "available", but in name only – loading a content reveals it's the Wide core instead. Everything was normal up to version 3.7, but that changed after I upgraded to version 4.0 using the " online updater" option.

- Which version of Lakka are you using?
  Version 4.0, aarch64.
- What system hardware are you using?
  Raspberry Pi 4.
- What did you do?
  Used the "online updater" to upgrade Lakka from version 3.7 to 4.0.
- What did you expect to happen?
  To keep the regular Genesis Plus GX core present in the OS.
- What happened instead?
  Regular Genesis Plus GX core was replaced by the Genesis Plus GX Wide core, even though all name references still point to the regular one.

ToKe79:
- update RetroArch to 1.10.2 and drop patch merged upstream
- RetroArch: disable core system files downloader (included in image)
- update almost all cores to their latest version
- drop duckstation core
- modify libretro_update.sh script to allow for updates to latest tagged commit
- modify some packages to use latest tag during automated update
- update liblcf to 0.7.0 (EasyRPG dependency) and drop patches
- do not rename `genesis_plus_gx_wide_libretro.so` to `genesis_plus_gx_libretro.so` in the install path

fixes #1667


lowburger:
RetroArch 1.10.2 already increase release commit twice, now up to 8b6fe0a, maybe it's final one :)
https://github.com/libretro/RetroArch/releases/tag/v1.10.2

Thx your great work in lakka project!

PR #1665, opened 2022-03-26
linux: enable CONFIG_HID_PID

dmrlawson:
This enables rumble on generic game controllers (e.g. raphnet adaptors)

Tested on Generic (x86_64) only, but should be generally applicable.

### DIFF
--- a/packages/linux/package.mk
+++ b/packages/linux/package.mk
@@ -308,6 +308,11 @@ pre_make_target() {
     ${PKG_BUILD}/scripts/config --set-str CONFIG_EXTRA_FIRMWARE_DIR "external-firmware"
   fi
 
+  # enable rumble for PID-compliant game controllers
+  if [ "${DISTRO}" = "Lakka" ] && [ ! "${LINUX}" = "L4T" ]; then
+    ${PKG_BUILD}/scripts/config --enable CONFIG_HID_PID
+  fi
+
   if [ ! "${LINUX}" = "L4T" ]; then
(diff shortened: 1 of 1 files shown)


GavinDarkglider:
Please dont over ride options for L4T builds, we use a much older kernel, and it can cause unforeseen issues, for no great reason.

Just drop your change below this line: https://github.com/libretro/Lakka-LibreELEC/pull/1665/commits/bec16e4af5547a8d8edf713759f49b470b385311#diff-ae40bca272825c429f33ecde254d7fa3779be3843fe25bd54954258bc2c4f16dR316

dmrlawson:
That's fair. 

I decided to just add a `[ ! "${LINUX}" = "L4T" ]` condition to separate it from that block below (which looks like the bit that does the work). What do you think?

I have used this on L4T, though not on Lakka or Switch (and I have no way to test or support it there), so I think it would likely be fine but totally understand not wanting to mess with it.

GavinDarkglider:
> That's fair.
> 
> I decided to just add a `[ ! "${LINUX}" = "L4T" ]` condition to separate it from that block below (which looks like the bit that does the work). What do you think?
> 
> I have used this on L4T, though not on Lakka or Switch (and I have no way to test or support it there), so I think it would likely be fine but totally understand not wanting to mess with it.

Yeah, I just dont want to mess with it, a ton of the other changes broke building, hince the skipping them, and pushing my own config.

PR #1684, opened 2022-04-13
Amlogic: disable composite output on tvboxes

### LINKED ISSUE #1659: Freeze on lakka logo on x96 mini using AMLGX aarch64 4.0 image
i bootable my sd card with etcher and plug sd card in my device boot successfully.  first step resize partition and reboot after 15 second is ok but after reboot device lock on lakka image and no happen anything.

- Which version of Lakka are you using? 4.0
- What system hardware are you using? x96mini 2g ram / 16g rom
- What did you do?  truning on device
- What did you expect to happen? boot to main menu lakka
- What happened instead? freeze at lakka logo

igyenes:
Disable composite output to force retroarch to start on HDMI.
Fixes #1659

### DIFF
--- a/projects/Amlogic/options
+++ b/projects/Amlogic/options
@@ -26,6 +26,11 @@
   # kernel serial console
     EXTRA_CMDLINE="systemd.debug_shell=ttyAML0 console=ttyAML0,115200n8 console=tty0"
 
+  # On tvbox force HDMI output by disabling composite
+    if [ "$UBOOT_SYSTEM" = "box" ]; then
+      EXTRA_CMDLINE+=" video=Composite-1:d"
+    fi
+
 ################################################################################
 # setup build defaults
 ################################################################################


ToKe79:
Thank you for your patch.

Do you have to build the image including `UBOOT_SYSTEM=box` in make options (e.g. `make image DISTRO=Lakka PROJECT=Amlogic DEVICE=AMLGX UBOOT_SYSTEM=box`, or is the `EXTRA_CMDLINE` also picked up when building without this ENV = all uboot-systems based on output of `scripts/uboot_helper`, i.e. `make image DISTRO=Lakka PROJECT=Amlogic DEVICE=AMLGX`?

We usually build all images for all targets - without adding `UBOOT_SYSTEM` to make options, which I think sources the project/device options only once and is not sourced again after setting `UBOOT_SYSTEM`. Therefore maybe it would be better to add the modification of the EXTRA_CMDLINE env before writing the `UEnv.txt`. I think we also were solving this issue on the Lakka-v3.x branch for RK3326 device where we needed to add rotation to the cmdline, so @davidgfnet added a helper function to the device options and this function is called before `UEnv.txt` is written.

https://github.com/libretro/Lakka-LibreELEC/blob/90c67195c03ac4ccbf5fcc861c9f252a4bd818ff/projects/Rockchip/devices/OdroidGoAdvance/options#L40-L46

https://github.com/libretro/Lakka-LibreELEC/blob/90c67195c03ac4ccbf5fcc861c9f252a4bd818ff/scripts/image#L369-L372

igyenes:
I have built for testing the changes on docker with:
`docker run --rm -v `pwd`:/build -w /build -it -e PROJECT=Amlogic -e DEVICE=AMLGX -e ARCH=aarch64 lakka make image`
So not passing `UBOOT_SYSTEM`.

The other way that I have also tried and working is add to a new `mkimage_uEnv-box()` part which contains `video=composite-1:d` to `/projects/Amlogic/bootloader/mkimage` and calling this function later on at case statement. 

 https://github.com/libretro/Lakka-LibreELEC/blob/11d43f248f2f2be214b995160e5c2fda54b42591/projects/Amlogic/bootloader/mkimage#L26-L31

https://github.com/libretro/Lakka-LibreELEC/blob/11d43f248f2f2be214b995160e5c2fda54b42591/projects/Amlogic/bootloader/mkimage#L60-L65

Honestly I do not know which is the best way out of the three, I assume the `mkimage` script runs recursively when creating the image, so this way is safe too.

PR #1687, opened 2022-04-20
Fix Lakka-LE-master Switch build

GavinDarkglider:
This has not been boot tested. Also, needs work with kernel headers for ffmpeg patches, and needs moonlight core added back.


GavinDarkglider:
This is tested working for the most part, I didnt test all of the cores, but vulkan, decoder, opengl, audio, dock, sleep, .... all work.

There are probably some cleanups that can be done, as this was kind of a quick and dirty make it build, and boot, and generally work to the same level as 4.x branch with as minimal modifications possible.

Most changes, are because of patch rebases, or because I pulled old version from 4.x tree to projects/L4T/packages.

ToKe79 (approved):

PR #1702, opened 2022-05-11
puae2021: add new libretro core

### LINKED ISSUE #1691: Include PUAE 2021 Core
Can the Commodore - Amiga (PUAE 2021) core be included in Lakka distros, especially for slower devices like ODroid and others? It is useful for games and cases where current PUAE struggles to run at full speed. PUAE 2021 seems is much faster than the current PUAE, and is already available in Retroarch on the Switch.

ToKe79:
puae based on older version for better performance.
closes #1691


gouchi:
LGTM

gouchi (approved):
LGTM

PR #1713, opened 2022-07-08
WIP - Add Support for  Ayn Odin

GavinDarkglider:
This is a WIP but mostly working, with the right modifications.

There are issues in this PR.

#1. nothing will come up on screen, since I didnt edit initramfs in this to patch out hiding console text(This will effect all boards). This is tied into, and possibly an issue caused by problem #2.
#2. X doesnt refresh screen properly especially if console text is missing(Modesetting driver issue) GL works, until you return to menu, then return to game, and Vulkan only works if you switch to VT, or put into sleep mode first.
#3. Sleep mode is broken in kernel(So I disabled it, until such time as it works.)
#4. USB mode switching doesnt work, once again a kernel issue, and battery charger turns off host mode power, so while usb host works, it currently only works with a powered USB hub.
#5. PD/Quick Charge(Kernel issues...)

What works without issues so far:
Controller, touchscreen, audio, battery stats/charger(non PD).

The first 2 issues are related to X, but I havnt tracked them down. KMS is not an option, as retroarch doesnt rotate screen properly at all. Also, the Odin supports a dock like the switch and when the time comes, wayland/kms do not support randr by default, X is about as light as you are going to get on an Arm64 based board with all of the support you need to make a dock work. Once again, this is future planning, as PD/DP isnt working in kernel yet.


dmrlawson:
Having tested a number of iterations of this, it's going really well!

One thing I'd like to discuss is the sd card layout.

I appreciate there may be reasons for putting everything in a single FAT32 partition on Switch, but afaik all other targets use the FAT32 + ext4 layout. My personal preference would be to stick with that, but I'm open to change if there's justification.

Otherwise this is as good as we can get for now as far as I'm aware. It and it does boot and games start. As mentioned in the summary there are various issues with X which I'm aware @GavinDarkglider has spent a significant amount of time trying to overcome. Hopefully we'll find a workaround in the not-to-distant future.

dmrlawson:
It might be worth adding it to the list of targets in `build_all.sh` and `pkg_all.sh`, not sure if that's sufficient to get the build server to build it

GavinDarkglider:
> Having tested a number of iterations of this, it's going really well!
> 
> One thing I'd like to discuss is the sd card layout.
> 
> I appreciate there may be reasons for putting everything in a single FAT32 partition on Switch, but afaik all other targets use the FAT32 + ext4 layout. My personal preference would be to stick with that, but I'm open to change if there's justification.
> 
> Otherwise this is as good as we can get for now as far as I'm aware. It and it does boot and games start. As mentioned in the summary there are various issues with X which I'm aware @GavinDarkglider has spent a significant amount of time trying to overcome. Hopefully we'll find a workaround in the not-to-distant future.

For simplicity I chose to go with fat32 layout. There is no need to flash an image with etcher, and it can be easily modified to boot on a batocera sdcard as well. It was also easier to patch the switch stuff, in the image creator to build this image, than it would be to build a full image, especially since root is always squashfs anyway. Also, we stop that script, before it really starts doing the pi image thing..... Personally I dont see any point in any build having an image if all it does is drop the squashfs on an ext4 partition, then do the same thing. Basically, this is easily installed with built in tools in most OS's, and doesnt need any special tools to flash SD card. On top of that, I would argue that the only time you should partition something like an sdcard with multiple partitions, it if it is completely unavoidable.

Also, as I stated in the PR, this wont even show retroarch in general, as I removed the hacky stuff from init script that effects all boards, but gets X to start and display, from the PR. So it is basically fully working and not working at all.....

GavinDarkglider:
> It might be worth adding it to the list of targets in `build_all.sh` and `pkg_all.sh`, not sure if that's sufficient to get the build server to build it

When the time comes, sure, but even then, vudiq would have to add the target to the build server scripts.

build_all.sh
```diff
@@ -83,6 +83,7 @@ targets="\
 	Allwinner|H6|aarch64|image \
 	Allwinner|R40|arm|image \
 	Amlogic|AMLGX|aarch64|image \
+        Ayn|Odin|aarch64|image \
```
gouchi, line 86: Remove space ?
```

diff --git a/build_all.sh b/build_all.sh
index a2fd7180bf..b50e119a1b 100755
--- a/build_all.sh
+++ b/build_all.sh
@@ -83,6 +83,7 @@ targets="\
 	Allwinner|H6|aarch64|image \
 	Allwinner|R40|arm|image \
 	Amlogic|AMLGX|aarch64|image \
+        Ayn|Odin|aarch64|image \
 	Generic||i386|image \
 	Generic||x86_64|image \
 	L4T|Switch|aarch64|image \
```

packages/lakka/retroarch_base/retroarch_joypad_autoconfig/joypad_configs/udev/Odin Gamepad.cfg
```diff
@@ -0,0 +1,34 @@
+input_driver = "udev"
+input_device = "Odin Gamepad"
+input_device_display_name = "Odin Gamepad"
+input_vendor_id = "8224"
+input_product_id = "273"
+
```
gouchi, line 6: We should add label and add it to [retroarch-joypad-autoconfig repository](https://github.com/libretro/retroarch-joypad-autoconfig) ?
[
Source Android file](https://github.com/libretro/retroarch-joypad-autoconfig/pull/847/files)

vanfanel:
X? Good I read this before buying the console for Lakka...
X stinks and adds a ton of input lag. Forget about the docking and go KMS/DRM.

Ntemis:
Unfortunately docking support is essential and yes if there were any other ways to support it without X that would be amazing but as far as we know there isn't yet.This doesn't mean that we cant switch to anything that works in the future though so have faith

packages/lakka/retroarch_base/retroarch_joypad_autoconfig/joypad_configs/udev/Odin Gamepad.cfg
GavinDarkglider: Dont disagree, but not sure I am ready to commit it yet, as the buttons on the back of the console arent included yet.

GavinDarkglider:
> X? Good I read this before buying the console for Lakka... X stinks and adds a ton of input lag. Forget about the docking and go KMS/DRM.

@vanfanel X isnt an issue for lag, and infact fixes a lot of things, such as touchscreen, and rotation, which cant be done without fixing retroarch rotation at the very least in KMS mode. As for the dock, the plan is to support it when the kernel can. You want that fixed, either get the weston devs to add support for randr extensions, and/or fix the issue in retroarch that breaks screen rotation. Then you still have to figure out how you are going to handle the touchscreen and the dock. Next time you want to complain about 3 weeks of my life, you get the system, and port it yourself, then you can make design plans/changes...... That being said, since I am one of the only people adding board support over standard libreelec tree boards, dont complain, or you can take my place figuring this out.

build_all.sh
GavinDarkglider: Hmmm..... Need to look into that, as it lined up in my editor.....

Ntemis:
@GavinDarkglider when this is ready for a review please remove the label

GavinDarkglider:
> @GavinDarkglider when this is ready for a review please remove the label

This has been ready for review for a while, excepting the issues with vulkan, which hopefully are fixed in mesa updates eventually. It could be merged now, and mostly works, I just dont like merging support for something that isnt fully supported.

I guess vulkan support could be disabled, but, it kind of defeats the purpose some, since things like dolphin are going to be crap compared to the stock android OS version.

PR #1717, opened 2022-07-21
Update orangepi-4-lts build

dmrlawson:
Image here: https://nightly.builds.lakka.tv/members/dmrlawson/Lakka-RK3399.aarch64-4.2-devel-20220717183455-62320b3-orangepi-4-lts.img.gz

Not sure why there are syntax errors with the dts file from armbian, I could use some help with that.

This still won't add wifi/ethernet/bluetooth support for now, but it's likely we can copy their patches for that too.

I split the makefile and dts patches to make updating easier. I'd be happy to rename them if someone has a suggestion.

I've tested that the above image boots to retroarch and that a usb gamepad works, I've not tested any cores yet.


Ntemis:
The ethernet and wifi chip are not supported by the mainline kernel we use atm so what you experience is normal

PR #1744, opened 2022-09-19
More build fixes for odin

dmrlawson:
Still doesn't get into retroarch.

This also creates a log file for Xorg. Might not want to merge with that - or maybe we do and we can clean it up later.


Ntemis:
yes since odin is in not so usable state atm lets remove it later

PR #958, opened 2020-05-30
libretro: update to latest core and latest RetroArch stable

gouchi:
Update core to latest and RetroArch to latest stable.

Most important changes : 

- easyrpg : add libfmt to package dependencies
- gcc :  add libatomic for kronos core
- liblcf : update needed for easyrpg and remove benchmark (not needed) as it failed
- opera: rename 4do core
- play! : remove std++ static build
- retroarch : update to version 1.8.8

PR #1746, opened 2022-09-27
gcc: enable libatomic

dmrlawson:
With this patch Odin will boot into retroarch, thanks to a lot of help from @GavinDarkglider 

I'm pretty sure we need it for other targets too

### DIFF
--- a/packages/lang/gcc/package.mk
+++ b/packages/lang/gcc/package.mk
@@ -14,14 +14,18 @@ PKG_DEPENDS_HOST="ccache:host autoconf:host binutils:host gmp:host mpfr:host mpc
 PKG_DEPENDS_INIT="toolchain"
 PKG_LONGDESC="This package contains the GNU Compiler Collection."
 
-case ${TARGET_ARCH} in
-  arm|riscv64)
-    OPTS_LIBATOMIC="--enable-libatomic"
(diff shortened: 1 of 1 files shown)


Ntemis:
I just disable this in my previous merge
https://github.com/libretro/Lakka-LibreELEC/commit/fd58bd5200cd8adfda2d08e73a38154654b91be9#diff-303be7bf01999d6462f59cd2749f00a21d98446e7023e109e82b531fe9ba9cceL82
The issue was alive before this, i find it kinda weird enabling libatomic back now fixes it!
Did you performed a clean build?

dmrlawson:
I'll try a clean build to be sure. I was getting this:
image
and this single change created a build that boots. Before that we were really struggling to see what was going wrong because X appeared to be starting fine.

We used to have it in v4: https://github.com/libretro/Lakka-LibreELEC/pull/958/files#diff-303be7bf01999d6462f59cd2749f00a21d98446e7023e109e82b531fe9ba9cceR69

The bit you removed does `--disable-libatomic -> --enable-atomic` which looks wrong, but I guess it gets rid of `--disable-libatomic` and it might be enabled by default?

Ntemis:
Thanks for the clear explanation

PR #1755, opened 2022-10-09
bcm2835-bootloader: sync URL and version with upstream

dmrlawson:
We got a 404 from github how it was and this will make future merges easier


dmrlawson:
Actually this fails - did LE block the build server from their mirror?

PR #1789, opened 2022-12-06
packages: add gamemode

nekoprog:
Add gamemode package, but it's not on lakka PKG_DEPENDS_TARGET because of DBUS error.

> Dec 06 10:07:27 Lakka systemd[1]: retroarch.service: Scheduled restart job, restart counter is at 41.
> Dec 06 10:07:27 Lakka systemd[1]: Stopped retroarch.service.
> Dec 06 10:07:27 Lakka systemd[1]: Starting retroarch.service...
> Dec 06 10:07:27 Lakka systemd[1]: Started retroarch.service.
> Dec 06 10:07:27 Lakka retroarch[770]: GameMode ERROR: Could not connect to bus: Using X11 for dbus-daemon autolaunch was disabled at compile time, set your DBUS_SESSION_BUS_ADDRESS instead
> Dec 06 10:07:27 Lakka retroarch[770]: dbus[770]: arguments to dbus_connection_unref() were incorrect, assertion "connection != NULL" failed in file /home/nekoprog/Lakka-LibreELEC/build.Lakka-RPi2.arm/build/dbus-1.14.0/dbus/dbus-connection.c line 2820.
> Dec 06 10:07:27 Lakka retroarch[770]: This is normally a bug in some application using the D-Bus library.
> Dec 06 10:07:27 Lakka retroarch[770]:   D-Bus not built with -rdynamic so unable to print a backtrace
> Dec 06 10:07:27 Lakka systemd[1]: retroarch.service: Main process exited, code=killed, status=6/ABRT
> Dec 06 10:07:27 Lakka systemd[1]: retroarch.service: Failed with result 'signal'.

Hoping someone might be able to solve the error.

### DIFF
--- a/packages/devel/inih/package.mk
+++ b/packages/devel/inih/package.mk
@@ -1,0 +1,8 @@
+PKG_NAME="inih"
+PKG_VERSION="54"
+PKG_SHA256="b5566af5203f8a49fda27f1b864c0c157987678ffbd183280e16124012869869"
+PKG_LICENSE="BSD"
+PKG_SITE="https://github.com/benhoyt/inih"
+PKG_URL="https://github.com/benhoyt/inih/archive/refs/tags/r${PKG_VERSION}.tar.gz"
+PKG_DEPENDS_TARGET="toolchain"
+PKG_LONGDESC="Simple .INI file parser in C, good for embedded systems"
--- a/packages/tools/gamemode/package.mk
+++ b/packages/tools/gamemode/package.mk
@@ -1,0 +1,9 @@
+PKG_NAME="gamemode"
+PKG_VERSION="1.7"
+PKG_SHA256="c1860f76f1d4c0d6e3965e52de21c824f24791049946da728da50f0c63748389"
+PKG_LICENSE="BSD"
+PKG_SITE="https://github.com/FeralInteractive/gamemode"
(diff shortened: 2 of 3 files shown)


gouchi:
Thank you for this update but as there is only one application running on Lakka I am not sure we should enable it.

I think we should [disable it](https://github.com/libretro/RetroArch/blob/master/retroarch.cfg#L1024) ? @Ntemis @GavinDarkglider

nekoprog:
I think we can make it work, if somehow host system can pass `DBUS_SESSION_BUS_ADDRESS` during bootstrapping to target build.

GavinDarkglider:
I see no point in including it, and actually plan on patching out the menu item in retroarch all together. We already give control of CPU governor/clocks. and with only one app, the only other thing game mode really does is swap resolutions and do ionice stuff so your game gets priority over other apps. The only other apps are daemons, so really no point. Also, I dont see a point in ionice on a build like this anyway, since we arent going to be doing a ton of read/writing with most emulation, it will get read into ram, so it is only loaded once, unless you are dealing with bigger roms, like ps2 iso's, since most boards have a decent ram capacity.

nekoprog:
Thank you for the clarification. I will close this PR.

PR #1790, opened 2022-12-07
add ScummVM mainline core package

spleen1981:
Adds ScummVM mainline libretro core to Lakka, built directly from untouched mainline ScummVM source.

As current official ScummVM libretro core is based on a very old fork of ScummVM, this core aims to be always in sync with ScummVM mainline repository, with a maintenance effort from low to none.

ScummVM main repo is included as a submodule. To sync the libretro core it should be sufficient to update the submodule and build

- No patch is applied to the submodule
- Submodule updates will be generally committed following ScummVM official releases (though it can be updated to any commit upstream, considering that adjustment to core Makefile and sources in src may be required to build)
- Datafiles and themes bundle (scummvm.zip) and core.info files are built automatically based on current submodule source

This core is supposed to be eventually moved to Libretro official repositories (pls let me know how in case) and superseed current outdated scummvm-libretro core.

### DIFF
--- a/distributions/Lakka/options
+++ b/distributions/Lakka/options
@@ -289,6 +289,7 @@
                   sameboy \
                   sameduck \
                   scummvm \
+                  scummvm_mainline \
                   snes9x \
                   snes9x2002 \
                   snes9x2005 \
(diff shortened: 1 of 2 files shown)


gouchi:
Thank you for your the update !

I think it will be better if you update the Libretro official repository.

You should try to connect if possible to [Discord](https://discord.com/invite/VZ2b7wghxR) to #programming.

Ntemis:
Did you test it on arm and aarch64 targets ?

spleen1981:
> Did you test it on arm and aarch64 targets ?

Yes, built and tested with arm, aarch64 and x86_64.

PR #1809, opened 2022-12-30
Lakka v4.x general

GavinDarkglider:
This PR does 2 general things....

It downgrades ppsspp to version from lakka v4.2, which actually loads/works.

It bumps Lakka version to 4.3

### DIFF
--- a/distributions/Lakka/version
+++ b/distributions/Lakka/version
@@ -2,4 +2,4 @@
   LIBREELEC_VERSION="devel"
 
 # OS_VERSION: OS Version
-  OS_VERSION="4.2"
+  OS_VERSION="4.3"


spleen1981:
Downgrading ppsspp should not be necessary with #1811, if it is confirmed that part should be removed

GavinDarkglider:
> Downgrading ppsspp should not be necessary with #1811, if it is confirmed that part should be removed

I am less concerned about fixing newer version of core, and more concerned about not breaking already supported cores...... if this your fix fixes it on my boards, great, if not, I would prefer to keep the working versions in stable tree.

Ntemis:
@GavinDarkglider ppsspp is now fixed thanks to @spleen1981 please revert ppsspp changes so i can merge, thanks

GavinDarkglider:
> @GavinDarkglider ppsspp is now fixed thanks to @spleen1981 please revert ppsspp changes so i can merge, thanks
I dropped my ppsspp changes, so this is just a version bump now.

PR #1819, opened 2023-01-11
spleen1981/scummvm_mainline to libretro/scummvm core (4.x)

spleen1981:
Move spleen1981/scummvm_mainline to libretro/scummvm and remove legacy core


packages/lakka/libretro_cores/scummvm/package.mk
```diff
@@ -1,37 +1,41 @@
...
-PKG_URL="${PKG_SITE}.git"
+PKG_URL="$PKG_SITE.git"
 PKG_DEPENDS_TARGET="toolchain"
-PKG_LONGDESC="ScummVM with libretro backend."
+PKG_SHORTDESC="The ScummVM adventure game engine ported to libretro."
+PKG_LONGDESC="ScummVM is a program which allows you to run certain classic graphical point-and-click adventure games, provided you already have their data files."
 PKG_TOOLCHAIN="make"
-
-PKG_MAKE_OPTS_TARGET="-C ../backends/platform/libretro/build/"
+PKG_LR_UPDATE_TAG="yes"
```
ToKe79, line 11: Do you tag "stable" releases in the repository? If so, then this is ok, otherwise we should update to latest commit - this line needs to be removed

spleen1981: yes, I'm using [tags ](https://github.com/libretro/scummvm/tags) to mark "stable" points

ToKe79 (approved):

PR #1829, opened 2023-02-12
optional exFAT partition

ToKe79:
based on feature request #646 

- optional exFAT partition created during first boot / partition setup
- bios files copied from exFAT to default location
- no change to default behavior


ToKe79:
obsolete/closed branch

PR #1830, opened 2023-02-12
optional exFAT partition

ToKe79:
based on feature request #646 

- optional exFAT partition created during first boot / partition setup
- bios files copied from exFAT to default location
- no change to default behavior


GavinDarkglider:
@ToKe79 Might have found a solution to this issue, which would also fix similar issues with this PR. https://github.com/libretro/Lakka-LibreELEC/issues/1922

If it works and can be implemented for switch, it should add fake permissions, similar to old umsdos patch set for running Linux from vfat. https://github.com/GregorR/upfs

Also, if it works, it should conclude the distro side of the switch development, outside of code cleanups, and RetroArch stuff that still needs to be done.

PR #1832, opened 2023-02-15
RPi4-PiBoyDmg: switch back to compiling xpi_gamecon as a module

dmrlawson:
This fixes PiBoy DMG not booting with Lakka v5.

It reverts https://github.com/libretro/Lakka-LibreELEC/commit/513da9989217535527c4444b1131a01e67340f18 and updates the module to the latest (which adds the `running` variable).

For whatever reason we just get a blank screen and the system appears to do nothing without this change (the data partition doesn't even get resized on first boot). With this change everything works as before, but we might have an issue with upgrading again.


ToKe79 (dismissed):
This looks good. Please finalize the PR.

dmrlawson:
I tested an update (after adding the second commit) and it worked without shutting off part way through. Not sure if this was just luck or the latest Experimental Pi firmware.

dmrlawson:
I don't think this will work as I think the update script (`init` in the busybox package) runs before the filesystem is mounted (though I may be mistaken).

PR #1843, opened 2023-04-15
Add support for RPi4 GPICase2

audunmg:
Hi, 

Here's a pull request to get the GPICase2 working, I've tested it on my own device.

Instead of using the safe shutdown script from [333mhz/GPiCase2-Script](https://github.com/333mhz/GPiCase2-Script) I copied the [packages/lakka/lakka_tools/gpicase_safeshutdown](https://github.com/libretro/Lakka-LibreELEC/tree/Lakka-v5.x/packages/lakka/lakka_tools/gpicase_safeshutdown) for the GPiCase 1 since it's doing almost exactly the same thing. Pins are the same, the only difference is the power on pin is held low instead of high. 

Maybe it should be merged into the same package? It might require a config file, or maybe it can be fixed in the build system? I don't know the best way to do that though. Suggestions welcome.

The switching between HDMI and LCD is also not working as in that script, since the original 333mhz/retroflag script switches between them by replacing config.txt and rebooting the whole thing with LCD disabled/enabled.

I strongly feel there is a better way to do that without a reboot, but haven't worked out exactly how yet, so that's why it's not included.

It's my first contribution, so this probably needs more work, suggestions on how to be better are welcome.


antonlacon:
> Instead of using the safe shutdown script from [333mhz/GPiCase2-Script](https://github.com/333mhz/GPiCase2-Script) I copied the [packages/lakka/lakka_tools/gpicase_safeshutdown](https://github.com/libretro/Lakka-LibreELEC/tree/Lakka-v5.x/packages/lakka/lakka_tools/gpicase_safeshutdown) for the GPiCase 1 since it's doing almost exactly the same thing. Pins are the same, the only difference is the power on pin is held low instead of high.
> 
> Maybe it should be merged into the same package? It might require a config file, or maybe it can be fixed in the build system? I don't know the best way to do that though. Suggestions welcome.

How's your python? You can combine the scripts by editing one to parse /etc/os-release for the device, and then set pull_up based on that.

audunmg:
That sounds good, I'll do that then.

packages/lakka/lakka_tools/gpicase_safeshutdown/scripts/safeshutdown_gpi.py
```diff
@@ -5,18 +5,27 @@
...
 power = LED(powerenPin)
 power.on()
 
+os_release_file = io.open("/etc/os-release")
+device = list(filter(lambda x: x.find('LIBREELEC_DEVICE') == 0, os_release_file.readlines() ))[0][18:-2]
+os_release_file.close()
+
+if device == "RPi4-GPICase2":
+    powerPinPullUp = False
+
```
antonlacon, line 21: [18:-2] is reducing `LIBREELEC_DEVICE="RPi4-GPICase2"` to `RPi4-GPICase`. Below should do what you want (double check, the hour is late):

```
with open('/etc/os-release', 'r') as data:
    content = data.read()
device = list(filter(lambda x: x.find('LIBREELEC_DEVICE') == 0, content.splitlines()))[0][18:-1]

powerPinPullUp = False if device == 'RPi4-GPICase2' else True
```

audunmg: Tested and working, thanks!

Ntemis:
good work champs

jayare5:
Whoa, GPi Case 2 was not supported yet? Time to get this device. :D

PR #1853, opened 2023-06-04
[WIP]Lakka 5.x Switch changes.....

GavinDarkglider:
This pretty much has everything on the lakka side finished, still needs proper decoder patches ported. I should probably adjust the  ppsspp ffmpeg script to actually work with all lakka builds, since ffmpeg 6.0 doesnt work for building against in latest ppsspp core, and the prebuilt versions from PPSSPP ffmpeg repo are broken at least on arm64. Sure there are other issues, I am having people beta test this now. I also still need to clean up/push the retroarch stuff upstream, so all of those patches are in the tree currently as well.


Ntemis (approved):

packages/multimedia/libva/package.mk
```diff
@@ -6,7 +6,6 @@ PKG_NAME="libva"
 PKG_VERSION="2.18.0"
 PKG_SHA256="9d666c70c12dfefcdd27ae7dea771557f75e24961d0ed4cb050d96fb6136f438"
 PKG_ARCH="x86_64"
-[ "${DISTRO}" = "Lakka" ] && PKG_ARCH+=" i386" || true
```
gouchi, line 9: @GavinDarkglider Why it has been removed ? Because we need to build libva for [i386](https://nightly.builds.lakka.tv/2023-10-05_5.x/0_logs/Generic.i386-2023-10-05_5.x.html) as it is required for [libva-utils](https://github.com/libretro/Lakka-LibreELEC/blob/Lakka-v5.x/packages/debug/libva-utils/package.mk).

PR #1919, opened 2024-01-07
iMX6: linux: config Bluetooth support and HIDP support as modules

by-gnome:
Since RF switch support is configured as a module, Bluetooth support and HIDP support must be configured as modules.
Make Bluetooth gamepad work again for IMX6.


Ntemis:
good find, thank you

PR #1920, opened 2024-01-09
vitaquake2: Add libretro core

synthic:
We already have Quake and Quake III Arena, why not Quake II? :)

Used vitaquake3 package file as a reference.


Ntemis:
thank you

PR #1926, opened 2024-01-29
Lakka v5.x switch 6

GavinDarkglider:
This is basically fixes issues in libreelec. That being said, Libreelec core packages in current upstream lakka tree are broken. Not an issue, as I have it working, and these were the fixes on the switch side, so I am pushing them. lol.


Ntemis:
Looks good to me, thanks

PR #1960, opened 2024-04-12
RPi: gamecon_gpio_rpi driver support

h1aji:
Updated gamecon_gpio_rpi driver to version 1.4 from https://github.com/marqs85/gamecon_gpio_rpi

### DIFF
--- a/packages/lakka/lakka_tools/gamecon_gpio_rpi/package.mk
+++ b/packages/lakka/lakka_tools/gamecon_gpio_rpi/package.mk
@@ -14,7 +14,8 @@ pre_make_target() {
 }
 
 make_target() {
-  make KDIR=$(kernel_path)
+  kernel_make KDIR=$(kernel_path) \
+        -C $(kernel_path) M=${PKG_BUILD} modules
 }
 
 makeinstall_target() {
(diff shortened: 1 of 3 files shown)


packages/lakka/lakka_tools/gamecon_gpio_rpi/package.mk
```diff
@@ -0,0 +1,23 @@
...
+PKG_LONGDESC="Gamecon GPIO rpi joystick driver"
+PKG_TOOLCHAIN="manual"
+PKG_IS_KERNEL_PKG="yes"
+
+pre_make_target() {
+  unset LDFLAGS
+}
+
+make_target() {
+  make KDIR=$(kernel_path)
```
ToKe79: this should be something like:
```
kernel_make -C $(kernel_path) M=${PKG_BUILD} modules
```

ToKe79 (changes requested):

packages/lakka/lakka_tools/gamecon_gpio_rpi/package.mk
h1aji:
```
make_target() {
  kernel_make KDIR=$(kernel_path) \
        -C $(kernel_path) M=${PKG_BUILD} modules
}
```

does that look correct?

ToKe79: I think KDIR is not needed, but it does not hurt.

ToKe79 (approved):

PR #1971, opened 2024-05-01
retroarch: fix the cheats directory mounting

RobLoach:
This updates the cheats directory to be referencing the installed directory from libretro-database.

### DIFF
--- a/packages/lakka/retroarch_base/retroarch/system.d/tmp-cheats.mount
+++ b/packages/lakka/retroarch_base/retroarch/system.d/tmp-cheats.mount
@@ -6,11 +6,10 @@ After=systemd-tmpfiles-setup.service
 DefaultDependencies=no
 
 [Mount]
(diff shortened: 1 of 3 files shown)


Ntemis:
thank you

ToKe79:
@RobLoach please can you then verify with next nightly build? I'm not sure if two overlays can share the same path. we mount `/usr/share/libretro-database` to `/tmp/database`. we set then content database to `/tmp/database/rdb`. therefore:
https://github.com/libretro/Lakka-LibreELEC/blob/5786c060f3888dae27fc8499df037cd3acd8b5a6/packages/lakka/retroarch_base/retroarch/package.mk#L192
should be changed to `/tmp/database/cht` and the `tmp-cheat.mount` units should be dropped completely. same for `/tmp/cheat` - should be dropped from tmpfiles.d

PR #1986, opened 2024-06-17
mupen64plus-next: Add Raspberry Pi 5 options

synthic:
Currently on the `RPi5` platform the available N64 emulation cores are crashing upon startup. This is due to a page size mismatch with the new Raspberry Pi 5 kernel, as outlined in issue #1982 

Since the upstream Makefile has no specific options for RPi5, I added the page size option and appropriate CPU optimizations to the compiler flags. Also, I added a case for compiling the `RPi5` platform as it was missing.

This fix still requires testing on a Raspberry Pi 5 device before merging.


synthic:
Tested on a Raspberry Pi 5 running Lakka v5.0 and mupen64plus-next launches games successfully.

Ntemis:
thank you

PR #1994, opened 2024-07-22
mupen64plus-next: Use upstream makefile

synthic:
Raspberry Pi 5 has finally been added to the makefile for mupen64plus-next, so we don't need to rely on a patch anymore. Also, the code has recently been rebased with tons of changes, so hopefully some of the issues with the core itself will be fixed in the newer version.


Ntemis:
I trust you tested  it. Let's merge for now as am vacationing.
And thanks 😊

synthic:
Yes, it builds successfully. Thanks! Have a great vacation! 🙂

PR #2074, opened 2025-05-28
Added VirtualXT core

andreas-jonsson:
Added VirtualXT core.

This core as been tested on RPi3/4 and x86_64 but should also run fine on any 32bit system as it has been built and tested in other CIs and projects.


Ntemis:
You should send a pr to add it to retroarch core database too

andreas-jonsson:
It is available in RetroArch already.

PR #2096, opened 2025-08-17
update openssl to 3.5.1 and openssh to 10.0p2

LordBubblesDev:
This pull request updates openssl and openssh. The packages are from LibreELEC directly, the only change applied was to make `/usr/bin/openssl-config` executable, which was necessary otherwise the OpenSSL configuration service fails to run.


GavinDarkglider:
This looks good to me.

Ntemis:
Thank you guys.

PR #2099, opened 2025-09-23
Fix #2098

### LINKED ISSUE #2098: Configuration in services section is not preserved after reboot
Lakka Version: H3.arm-5.0

Hello,

I try to turn off SAMBA by setting 

Services > SAMBA to OFF

It works fine but after rebooting it will always set to ON

Other settings in this section like SSH on/off don’t have this problem and is well preserved after reboot.

I noticed that this setting (and other settings like SSH in services section) is not save in the /storage/.config/retroarch/retroarch.cfg file like other “normal“ settings.

Maybe there is a place that I can set the SAMBA default value to OFF? That will do the job.

Thanks!

ckwun:
Toggle SAMBA on/off in menu will only result in create/remove of file `/storage/.cache/services/samba.conf`
Use this file to determine the current state.

### DIFF
--- a/packages/network/samba/scripts/samba-config
+++ b/packages/network/samba/scripts/samba-config
@@ -45,11 +45,11 @@ fi
 
 echo >>${SMB_TMP}
 
-if [ ! -f /storage/.cache/services/samba.disabled ]; then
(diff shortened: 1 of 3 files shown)


Ntemis (approved):
Thank you

PR #2188, opened 2026-03-20
addons/sundtek-mediatv: Fix error handling in the driver update script

gkapetanakis:
`packages/addons/driver/sundtek-mediatv/source/bin/sundtek-update-driver.sh` enables `set -e` which causes the script to exit on any non-0 exit status. At the same time it tries to check `$?` which can only be 0 since any non-0 status prevents the script from reaching that statement. This PR fixes that issue and enables the original error handling functionality to work.


ToKe79:
please, take this to upstream - https://github.com/LibreELEC/LibreELEC.tv
thank you!

PR #2196, opened 2026-04-09
pd777: update to 1.01.01

mittonk:
pd777: update to 1.01.01 for color fixes.

Tested on Raspberry Pi 4, 64-bit.


gouchi:
@Ntemis @ToKe79 LGTM.

Issue #2219, opened 2026-05-14
[Feature Request] Add pipewire by default for bluetooth audio


knightpp:
Is there a reason to not include pipewire or pulseaudio by default?

I've just built a custom image with pipewire and it works on my SBC.

ShigeakiAsai:
I tried this on an RPi5.
a) Changes made in pull request #2220
b) Added PIPEWIRE_SUPPORT="yes" to RPi5 options
c) Added --enable-pipewire to retroarch build parameters
After building, I connected a JBL Go4 via Bluetooth and got sound.
I felt there was some lag typical of Bluetooth. (I didn't change any retroarch settings.)

https://github.com/ShigeakiAsai/Lakka-LibreELEC/releases/tag/trial-RPi5-pipewire-support-001

Thanks
ASAI, Shigeaki

GavinDarkglider:
I have been using pulse audio for this, and dock switching for quite some time.

PR #2220, opened 2026-05-14
Pull pipewire, wireplumber updates from LibreELEC

knightpp:
I've copied over some audio related package updates. Also edited `show_config` to display pipewire support.

I could try to cherry-pick instead, but pipewire was not updated in a while.

### DIFF
--- a/config/show_config
+++ b/config/show_config
@@ -47,6 +47,7 @@ show_config() {
 
   config_message+="\n - ALSA support:\t\t\t ${ALSA_SUPPORT}"
   config_message+="\n - Pulseaudio support:\t\t\t ${PULSEAUDIO_SUPPORT}"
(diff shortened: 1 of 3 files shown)


gouchi:
Fix #2219

knightpp:
> Fix #2219

This PR does not address #2219. For that, I'd like to enable this option https://github.com/libretro/Lakka-LibreELEC/blob/afc4ca1dbdd0b0b68f99024f76a4f73c582bfdac/distributions/Lakka/options#L105

But I don't know if it's okay to do that distribution wide.

gouchi:
@knightpp Thank you, indeed I should have said related instead of fix.

@ToKe79 @Ntemis @GavinDarkglider Should we enable PipeWire ?

PR #2223, opened 2026-05-27
tm16xx: add patch file for fix display.service failure on Amlogic project

ShigeakiAsai:
## This PR fixes the "display.service" failure on Amlogic devices

The "display.service" is contained in the tm16xx linux driver package,
and only PROJECT=Amlogic uses this package.

This package's repository has fix commit.
https://github.com/jefflessard/tm16xx-display/commit/1b380d7d36958df5f5dcca8b4162394bb1490c67
So we use this commit as patch file.

### Build
I confirmed PROJECT=Amlogic build is passed with this PR's patch

### Test
I confirmed the "display.service" is not started on my khadas-vim3l.
(TM16xx Display device is not connected)

```
###########################################
# Lakka - The DIY retro emulation console #
# ...... visit http://www.lakka.tv ...... #
###########################################

Lakka (community): devel-20260527200956-afc4ca1 (AMLGX.aarch64)
Lakka:~ # journalctl | grep -i display.service
May 27 11:39:49 Lakka systemd[1]: display.service was skipped because of an unmet condition check (ConditionPathIsDirectory=/sys/class/leds/display).
Lakka:~ #
```

Thanks
ASAI, Shigeaki

### DIFF
--- a/packages/linux-drivers/tm16xx/patches/tm16xx-0001-display_service-tweaks-to-align-with-upstream-use.patch
+++ b/packages/linux-drivers/tm16xx/patches/tm16xx-0001-display_service-tweaks-to-align-with-upstream-use.patch
@@ -1,0 +1,43 @@
+From 1b380d7d36958df5f5dcca8b4162394bb1490c67 Mon Sep 17 00:00:00 2001
+From: Christian Hewitt <christianshewitt@gmail.com>
+Date: Tue, 9 Sep 2025 08:10:35 +0000
+Subject: [PATCH] display.service: tweaks to align with upstream use
+
+Set 'After' and 'Wants' conditions so the service does not
+run until after `time-sync.target` so the clock displayed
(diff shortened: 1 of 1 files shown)


Ntemis:
why you just didn't update the package version?

ShigeakiAsai:
I'm sorry for the lack of information.

There are 2 reason.

a) upstream LibreELEC 12.2 uses this package version.
  https://github.com/LibreELEC/LibreELEC.tv/blob/libreelec-12.2/packages/linux-drivers/tm16xx/package.mk#L5

b) If only package version bumps, shows build error.
  upstream LibreELEC 13/master uses more updated version,
  but it needs many packages changing.
  https://github.com/LibreELEC/LibreELEC.tv/pull/10478

Therefore, I thought better to use a fix patch file as a solution to this problem.

PR #2228, opened 2026-06-13
Update to SDL3

RobLoach:
This is some initial work to move to SDL3 for input. Will require the SDL3 Pull Request:
https://github.com/libretro/RetroArch/pull/18833


Ntemis:
@RobLoach remove draft flag so we can merge it and thanks for your work.

RobLoach:
Do note that I'm not 100% sure if the dependencies are correct for this, or if the CMake build works. Also, looking to add the other drivers sometime too :shrug:

Ntemis:
> Do note that I'm not 100% sure if the dependencies are correct for this, or if the CMake build works. Also, looking to add the other drivers sometime too 🤷

Then is best if you push retroarch update here too so we get on par with upstream.

RobLoach:
Agreed. While it's in RetroArch proper, I'm not sure of the best time to update to it in Lakka.

RobLoach:
Going to hold in draft until some of the other features get in, primarily video and audio.

RobLoach:
While the sdl3 package should be good to go, we'll still need to update RetroArch to the latest.

Ntemis:
You can add it here or sent a diff one for this task

RobLoach:
@Ntemis Makes sense to me. Folded into this one.

PR #2241, opened 2026-08-26
Lakka-v6.1: Fix stale RetroArch/RetroArch-assets patches and SDL3 X11 dependencies

ShigeakiAsai:
## Summary

Fixes several build failures found in Lakka-v6.1's nightly build on 2026-08-25 that affect all devices — most stemmed from Lakka-side patches that had already been merged upstream (RetroArch, RetroArch-assets) and were failing to re-apply after a version bump, plus one genuine upstream RetroArch bug and one Lakka-side dependency resolution issue (SDL3 picking up a Chrome-addon-private `libXcursor`).

### retroarch_assets (1 commit)

- **Revert "retroarch_assets: add pkg/osd-font.ttf patch that is used by Lakka matrix screen saver"**
  The Katakana additions this patch made to `osd-font.ttf` are now merged upstream directly, so the binary patch fails to re-apply and breaks every package depending on `retroarch_assets`.

### retroarch (5 commits)

- **Revert "RPi3-Composite: RPi4-Composite: Add support RetroFlag cases"**
  These local patches (`001`/`002`/`003_retroarch-retroflag_picase_safeshutdown_*`) are now fully merged upstream (`Makefile.common`, `configuration.c/.h`, `lakka.h`, `menu_cbs_sublabel.c`, `menu_setting.c`, `msg_hash*.h`), even extended with `HAVE_RETROFLAG_RPI5`. Reapplying them fails; removed as no longer needed.

- **Rebase `retroarch-1002-add_date_and_time_menu.patch`**
  Upstream refactored the `LOCALAP_ENABLE`/`TIMEZONE` menu area (new `HAVE_RETROFLAG` block inserted, direct `change_handler`/`action_ok` struct assignment replaced by the `SETTINGS_ACTION_SET` macro), which broke this patch's context. Rebased with no functional change.

- **RPi3/4/5 and each Composite: Remove upstreamed retroflag_picase safeshutdown patches**
  Same root cause as the revert above — these RetroArch-side patches are already merged upstream and fail to reapply.

- **RPi3/4/5-Composite: Rebase `retroarch-use_notification_font_for_cheevos.patch`**
  `font_driver_get_message_width()`'s second argument changed upstream from a hardcoded `0` to `strlen(...)`, breaking this patch's context in `gfx_widget_achievement_popup_start()`. Rebased to keep the `strlen()` call while still switching to `gfx_widget_fonts.msg_queue`.

- **Fix unbalanced braces from upstream commit 20a9c21 (HAVE_LAKKA_SWITCH)**
  RetroArch's own C90-scoping fix ([libretro/RetroArch@20a9c21](https://github.com/libretro/RetroArch/commit/20a9c21ee91e3b6ee53c207671195ffe3ba4f9cc)) left a stray `}` after the `lakka-switch.h` include with no matching `{`, and a missing closing `}` before the `switch_oc`/`switch_cec`/`bluetooth_ertm` `#endif` — breaking any `HAVE_LAKKA_SWITCH=1` build (`PROJECT=L4T DEVICE=Switch`) with cascading "invalid storage class" errors. Patch rebalances both.
  Upstream PR planned.

### libXcursor_lakka (1 commit, new package)

- **Add system-shared libXcursor for Lakka (needed by SDL3)**
  This fork's only `libXcursor` package lives under `addon-depends/chrome-depends/libXcursor` and is deliberately built with `PKG_BUILD_FLAGS="-sysroot"` so it stays private to the Chrome addon. Packages that need a real system-wide Xcursor (e.g. SDL3) were silently resolving to that private copy and never seeing it in the shared toolchain sysroot. Adds a proper shared package instead, matching `libXfixes`/`libXrender`'s build flags and `libtool_remove_rpath` `post_configure_target` convention.

### SDL3 (1 commit)

- **Replace `libXcursor` with new `libXcursor_lakka` and add `libXtst` to X11 dependencies**
  SDL3 requires Xcursor and Xtst, but `libXcursor` resolved to the Chrome-addon-private package (never synced to the shared toolchain sysroot), and `libXtst` was simply missing from the `x11` `PKG_DEPENDS_TARGET` list. Fixes SDL3 configure failures on Generic/x11 and Ayn/Odin.

### Confirmation
- All projects/devices build were done with no error.
  > All successful!
  > :-)

  Base Lakka-v6.1 commit: 074b462890102a34fa7ac57731427d8c7ac3c8dc

  The following projects are passed.
  \-  Allwinner
  \-  Amlogic
  \-  Ayn
  \-  Generic
  \-  L4T
  \-  NXP
  \-  RPi
  \-  Rockchip
  \-  Samsung

- just only check build is passed

## Note
Investigation and patch drafting assisted by Claude (Anthropic).

Thanks
ASAI, Shigeaki


packages/lakka/lakka_depends/SDL3/package.mk
```diff
@@ -49,7 +49,7 @@ fi
 
 # Video
 if [ "${DISPLAYSERVER}" = "x11" ]; then
-  PKG_DEPENDS_TARGET+=" libX11 libXcursor libXext libXfixes libXi libXrandr libXrender"
+  PKG_DEPENDS_TARGET+=" libX11 libXcursor_lakka libXext libXfixes libXi libXrandr libXrender libXtst"
```
RobLoach, line 52: Thanks for the SDL3 fix.

Ntemis: But i see an issue here , is preferable to use mainstream names vs our own as these mistakes can easily happen in the future so @ShigeakiAsai if any way to make it work with libXcursor so no suffix needed we should do it.
Thank you

ShigeakiAsai: I think @Ntemis makes a valid point.
Could I have a little time to consider this?
The original libXcursor package is located at `packages/addons/addon-depends/chrome-depends/libXcursor`; the reason I created `libXcursor_lakka` was that I was unsure if it was appropriate to use a package from `chrome-depends`.
Since Lakka does not include Chrome, I will look into modifying the setup to make effective use of the existing package.
Many thanks!!

ShigeakiAsai: https://github.com/libretro/Lakka-LibreELEC/pull/2244
I'll mark this as ready for review once the remaining tests finish.

PR #2244, opened 2026-09-07
Lakka-v6.1: libXcursor: build the upstream package directly instead of a Lakka-specific fork

ShigeakiAsai:
## Summary

Follow-up to #2241. As raised in review, `libXcursor_lakka` duplicated
`packages/addons/addon-depends/chrome-depends/libXcursor` under a
Lakka-specific name instead of using the upstream package name
directly, which risks the same kind of stale-patch/naming mismatch
that PR was fixing in the first place.

`chrome-depends/libXcursor` is deliberately built with
`PKG_BUILD_FLAGS="-sysroot"` so it stays private to the Chrome addon's
build and is never installed into the shared toolchain sysroot. Lakka
doesn't build the Chrome addon at all, so that isolation serves no
purpose here - but SDL3 on X11 needs a real, shared `libXcursor`
visible in the sysroot.

## What this does

- `packages/addons/addon-depends/chrome-depends/libXcursor/package.mk`
  On `DISTRO=Lakka`, build without `-sysroot` (a normal shared
  package) instead of the Chrome-only private copy. LibreELEC itself
  is untouched (the `else` branch keeps `-sysroot`).
- Revert "libXcursor_lakka: Add system-shared libXcursor for Lakka
  (needed by SDL3)" - no longer needed once the package above is
  built as a normal shared package.
- `packages/lakka/lakka_depends/SDL3/package.mk`
  Depend on `libXcursor` (upstream name) instead of `libXcursor_lakka`.

## Testing

The following 3 devices with `DISPLAYSERVER="x11"` were clean-built
successfully together with PR#2243:

- Generic/x11.x86_64
- Ayn/Odin.aarch64
- L4T/Switch.aarch64

## Note

Investigation and patch drafting assisted by Claude (Anthropic).

Thanks
Asai, Shigeaki


GavinDarkglider:
@ShigeakiAsai I have disabled the need for sdl3 on the switch(in my builds, it isn't needed for pretty much everything I have tested. It also adds broken video option on switch to retroarch.)

As for the Odin build, it needs a ton of work, and at some point I will get to it, but it is a low priority for me currently. Tons of kernel/mesa work to be viable as an everyday usable thing. I have updated kernel changes somewhere from @Ntemis but I haven't had time to review them properly as finishing up the switch stuff has been fairly time consuming, and I still have tons to do.

